PR #64, opened 2025-09-15
Fileio

tdpham2:
- Move all tool calls to file io operation to avoid large outputs (e.g coordinates, vibrational frequencies) in LLMs' inputs/outputs.


tdpham2:
run_ase still returns the full vibrational frequency list when driver is "vib", "ir", or "thermo".

PR #70, opened 2025-11-20
Major Update: MCP Integration, ToolNode Refactor, and Notebook Refresh

tdpham2:
- Updated dependencies and removed most version locks
- Refactored the multi-agent graph to use LangGraph’s prebuilt ToolNode.
- Added full MCP capabilities and unified LangChain and MCP tools under a single interface for both single and multi-agent
- Updated example notebooks: Demo_single_agent, Demo_multi_agent.
- Added a new MCP notebook: Demo_MCP.


keceli:
I guess it is ok to release version locks in the dev, but we better keep them to be reproducible in the main branch.

PR #85, opened 2026-02-09
Fix #84 Gemini API error: Add proper type annotation for dipole_value field

### LINKED ISSUE #84: [DEV] Gemini API Error: Invalid GenerateContentRequest.tools parameter with dipole_value field
## Description
When using Gemini models via the Streamlit web app, queries fail with the following error:

Invalid argument provided to Gemini: 400 * GenerateContentRequest.tools[0].function_declarations[0].parameters.properties[ase_output].properties[dipole_value].items: missing field.

## Steps to Reproduce
1. Configure ChemGraph with a Gemini API key
2. Launch the Streamlit web app
3. Submit a query that triggers the ASE output schema
4. Observe the error

## Expected Behavior
Queries should complete successfully without schema validation errors from Gemini's API.

## Actual Behavior
The Gemini API rejects the function_declarations with a missing field error related to `dipole_value.items`.

## Environment
- Model: gemini-2.5-flash (from config.toml)
- API: Gemini
- Interface: Streamlit web app

arkaslittlemind:
## Problem
When using Gemini models via the Streamlit web app, queries fail with the error:

Invalid argument provided to Gemini: 400 * GenerateContentRequest.tools[0].function_declarations[0].parameters.properties[ase_output].properties[dipole_value].items: missing field.

The issue was that `dipole_value` in `ASEOutputSchema` was typed as `list` without specifying the item type. Gemini's API requires arrays to have an `items` property that defines the element type.

## Solution
Changed `dipole_value: list` to `dipole_value: List[Optional[float]]` in `src/chemgraph/models/ase_input.py`. This ensures Pydantic generates the correct JSON schema with the `items` field that Gemini expects.

## Testing
- ✅ Verified the fix resolves the Gemini API error in Streamlit web app
- ✅ Confirmed queries now work correctly with Gemini models
- ✅ No breaking changes - the field still accepts the same data types

### DIFF
--- a/src/chemgraph/models/ase_input.py
+++ b/src/chemgraph/models/ase_input.py
@@ -175,7 +175,7 @@ class ASEOutputSchema(BaseModel):
         default=None, description="Single-point energy/Potential energy"
     )
     energy_unit: str = Field(default="eV", description="The unit of the energy reported.")
-    dipole_value: list = Field(
+    dipole_value: List[Optional[float]] = Field(
         default=[None, None, None],
         description="The value of the dipole moment reported.",
     )


keceli (approved):
Thank you @arkaslittlemind for the fix.

PR #107, opened 2026-03-26
XANES support

vitorgrizzi:
Add XANES calculation tool. The user MUST install via pip:

mp_api 
parsl

And also have FDMNES installed. 

Then, the user must config parsl by changing internal variables in `run_fdmnes_parsl_workflow` function inside xanes_tools.py. This is for dev purposes, before merging to main we must find a better way to change these user-dependent variables such as FDMNES .exe path, walltime, account to charge, etc.


tdpham2:
@vitorgrizzi I made some significant changes for the tools to make sure those are compatible with the existing ChemGraph (schemas, Parsl, etc.) and added an MCP server for the XANES tools. I will test this on Aurora and Polaris before merging.

tdpham2:
Also I removed the multi-agent for now since we will add a separate multi-agent with RAG for XANES later on.

PR #127, opened 2026-05-21
Modified the EL backend implemenations, and added a EL backend test

harikrishna1410:
Modification to EL backend and added tests.


tdpham2:
Thanks @harikrishna1410. Looks good to me! I will merge it and update it with some recent changes on my end with Globus.

PR #134, opened 2026-06-18
fix(academy,agent): PR-120 follow-up — optional-dep contract + calculator wiring

JinchuLi2002:
Two fixes from PR-120 review feedback

### P1 — `[academy]` optional-dep contract restored
`chemgraph.academy.__init__` and `chemgraph.academy.core.__init__` were eagerly
importing `ChemGraphLogicalAgent`, which pulls in `academy.agent`. Any code path
that touched the package (including `chemgraph.cli.trace`'s `--trace-dir` flow
and pytest collection) crashed on a checkout without the `[academy]` extra,
even though the rest of the package (campaign spec, prompt profile, event log)
is pure stdlib + pydantic.

* Both `__init__`s grow a `__getattr__`-based lazy resolver for
  `ChemGraphLogicalAgent`. Pure modules stay eager so their 11-symbol surface
  is reachable without the extra. Accessing the lazy symbol without
  `academy-py` installed raises `ImportError` with an actionable
  `pip install 'chemgraph[academy]'` hint.
* 9 academy test modules gain `pytest.importorskip("academy")` so they skip
  cleanly instead of erroring at collection time when the extra is absent.

### P2 — calculator wiring regression restored
`f1593ab` ("revert: restore llm_agent.py to pre-academy shape") was meant to
drop event-callback wiring and turn-primitive imports, but its rewrite also
deleted the calculator-availability injection that was already in the file at
the dev-globus fork point. The deletion was a merge-conflict casualty:
`get_calculator_selection_context()` in `schemas/ase_input.py` has been dead
code on dev-globus since, and the two
`test_single_agent_initialization_injects_calculator_availability` regression
tests have been failing.

Restored the 28-line wiring block verbatim from `dev:llm_agent.py:464-490`
plus the matching `chemgraph.schemas.ase_input` import.

### Validation
- Without `academy-py`: `chemgraph.academy` imports cleanly, all 11 eager
  exports resolve, `cli.trace` loads, all 9 guarded test modules skip
  cleanly (`pytest.importorskip` triggers at line :12-:11).
- With `academy-py`: 48/48 academy tests pass.
- Both calculator-injection regression tests now pass.
- The 8 other test failures in the full suite are pre-existing on
  `dev-globus`, confirmed by stashing these changes and re-running. Not
  introduced here.

### DIFF
--- a/src/chemgraph/academy/__init__.py
+++ b/src/chemgraph/academy/__init__.py
@@ -1,14 +1,33 @@
 """Academy Agents integration for ChemGraph.
 
-Provides agent classes and utilities for deploying ChemGraph workflows
-across federated HPC infrastructure using the Academy framework.
+Public re-exports come in two tiers so the package honours the
+``[academy]`` optional-dep contract:
 
-Requires the ``academy`` optional extra.
+* **Eager** (pure stdlib + pydantic, always importable):
+  ``ChemGraphAgentSpec``, ``ChemGraphCampaign``,
+  ``ChemGraphDaemonConfig``, ``MCPServerSpec``, ``ResourceSpec``,
+  ``load_campaign``, ``resolve_campaign_resources``,
+  ``PromptProfile``, ``load_prompt_profile``, ``CampaignEvent``,
+  ``EventLog``. These let the dashboard, ``--trace-dir``, and the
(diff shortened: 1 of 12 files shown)


tdpham2:
LGTM. Thanks @JinchuLi2002

PR #136, opened 2026-06-22
feat(academy): federated cross-HPC campaigns via HTTP exchange + multi-site dashboard

JinchuLi2002:
## Summary

Adds federated cross-HPC ChemGraph Academy campaigns. Two agents on
different HPCs discover each other via deterministic UIDs over
Academy's HTTP exchange, exchange messages, and render in one merged
dashboard driven from the operator's Mac.

End-to-end verified on Aurora + Crux: `federated-chat` bounced a
counter 1→10 between sites, ~230 events streamed live, zero errors.

## How it works

**Before this PR**: campaigns were single-machine. The daemon used
Redis (started by rank 0) as the exchange, peer registrations were
written to a shared-FS file the other ranks polled, and rank 0
dispatched the kickoff message inline at the end of startup. Every
piece of that — Redis subprocess, shared FS, inline kickoff — assumes
one allocation on one machine. To cross HPCs, all three had to change:
the exchange must be reachable from both sites' compute nodes (no
shared Redis), peer rendezvous must happen without a shared FS, and
kickoff must wait until every site has come up.

**Identity (no shared FS)**: each rank computes peer UIDs locally as
`uuid5(NS, "{run_id}/{agent_name}")`. Both sites derive the same UID
for each agent without any network lookup. `discover()` is used as a
UID-keyed liveness probe — not for name resolution, since the hosted
exchange strips names from discovery responses.

**Exchange**: `--exchange-type http` targets Academy's hosted exchange
(Globus-auth'd) instead of a per-allocation Redis. Both sites' compute
nodes talk to the same public endpoint, so messages cross HPC
boundaries without any direct site-to-site network path.

**Bootstrap**: in single-machine campaigns rank 0 dispatches the
kickoff message inline. In federated mode every spawn-site passes
`--no-bootstrap`; the operator runs `chemgraph academy bootstrap`
once all sites have registered, which sends the kickoff to
`initial_agent` over the same HTTP exchange.

**Dashboard**: `--system aurora,crux` brings up one SSH ControlMaster
+ UAN relay + rsync mirror per site, then serves a merged event
view. Per-site `events.jsonl` are interleaved by timestamp; per-site
`status`/`placement` are merged to top-level keys so existing
single-site renderers work unchanged.

See `examples/academy/federated-chat/e2e_guide.md` for the full
four-terminal walkthrough.

## What's new

- **Cross-site identity** via deterministic UIDs.
- **`chemgraph academy bootstrap`** standalone subcommand.
- **`--exchange-type http`** with proxy passthrough through
  `mpiexec --genv` so MPI ranks can reach the public exchange.
- **Multi-site dashboard** (`--system aurora,crux`).
- **Self-cleaning UAN relay** with SIGTERM handlers and per-UAN
  process scan (handles Aurora's uan-0001..0010 round-robin alias).
- **Operator-visible lifecycle prints** in daemon + agent.
- **`federated-chat` packaged campaign** + e2e guide.

## Compatibility

Single-machine `run-compute` flow is byte-identical to pre-PR
behavior. Federated paths are gated on `--agents` / `--no-bootstrap`
/ `--exchange-type http` being explicitly set.

## Test plan

- [x] End-to-end `federated-chat` Aurora ↔ Crux (counter 1→10, ~230 events)
- [x] +770 LoC tests covering exchange dispatch, bootstrap, compute
      launcher, dashboard, deterministic UID properties


tdpham2:
@JinchuLi2002 Should I review this PR, or you are still making changes to this?

PR #143, opened 2026-07-09
Fix Streamlit UI bugs: render crash, IR modes, math heuristic, config…

keceli:
Fixes four bugs found while auditing the Streamlit UI, plus unifies on a single config.toml.

1. Chat-render crash (message_utils.extract_molecular_structure) A message whose JSON "answer" was a string or list that merely mentioned the words "numbers"/"positions" hit `"numbers" in answer` (a substring check) and then `answer["numbers"]`, raising an uncaught TypeError that crashed the entire conversation render. Now structure data must be a dict and the values must be lists; TypeError is caught.

2. Linear-molecule IR modes (_pages/main_interface._render_ir_spectrum) The frequency table hardcoded `df.iloc[6:]` to drop translational and rotational modes. Linear molecules (CO2, HCN, diatomics) have only 5 such modes, so one genuine vibration was silently dropped. Added _num_nonvibrational_modes()/_is_linear_geometry() to detect linearity from the run's structure (5 vs 6), defaulting to 6 when unknown.

3. Parenthetical math false positives (message_utils) The inline-math heuristic treated a bare "_" or "^" as math, so prose like "(a_1 symmetry)" or "(k_B T)" was mangled into broken KaTeX. Conversion now requires a LaTeX command, grouping braces, or a sub/superscript alongside "=" (a real equation). "(E = mc^2)" and "(\mathrm{...})" still convert.

4. CWD-dependent config path (ui/config.py) + single config.toml load_config/save_config defaulted to a cwd-relative "config.toml", so launching Streamlit from any other directory silently created and used a throw-away default. They now anchor to the repo-root config.toml. Removed the orphaned src/ui/config.toml (it was never loaded), so the app and CLI share one config.toml.

Adds regression tests for all four; UI suite passes 22/22.

### DIFF
--- a/src/ui/_pages/main_interface.py
+++ b/src/ui/_pages/main_interface.py
@@ -921,6 +921,65 @@ def _resolve_artifact_path(filename: str, directory: Optional[str]) -> str:
     return filename
 
 
+def _is_linear_geometry(atoms) -> bool:
+    """Return whether an ASE ``Atoms`` object is (near-)linear.
+
+    A linear molecule has one vanishing principal moment of inertia.
+
+    Parameters
+    ----------
+    atoms : ase.Atoms
+        Structure to inspect.
+
+    Returns
+    -------
+    bool
+        ``True`` when the geometry is linear.
+    """
+    try:
+        moments = sorted(abs(float(m)) for m in atoms.get_moments_of_inertia())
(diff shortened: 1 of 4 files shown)


tdpham2 (approved):
Look good to me!

Issue #135, opened 2026-06-19
Collaboration proposal: workflow reliability fixes and organic-molecule focused path


rockyzl:
Hi ChemGraph maintainers,

I have been testing ChemGraph locally and building workflow/UI reliability improvements in a fork. I would like to contribute upstream, but I noticed my fork currently contains a larger set of workflow changes, so opening one direct PR against `main` would be too broad and would include unrelated history.

Current fork branch with the latest local fix:
https://github.com/rockyzl/chemgraph-workflow-improvements/tree/fix/structured-results-dipole-routing

Recent fixes I tested locally include:
- target-scoped follow-up routing so a new query such as `glucose dipolemoment` does not reuse old reaction species from a previous methane combustion task;
- deterministic completion validation before final scientific answers, so incomplete tool evidence does not produce a fabricated numeric result;
- structured output support for multiple related scalar values, e.g. reaction enthalpy and reaction Gibbs free energy shown together when both are available;
- clearer persisted UI error/empty-response states;
- calculator policy hints for molecule properties such as dipole/IR/thermo to prefer TBLite/GFN2-xTB when the user did not specify another calculator.

Local verification on my fork: `299 passed, 6 skipped`.

I would like to collaborate in the way that is most useful for the project. Would you prefer that I:

1. split these into small PRs against upstream `main`, starting with the smallest validation/routing fix;
2. open a draft PR from the broader fork branch for design review first;
3. focus first on an organic-molecule workflow/profile version, with TBLite-first defaults and stricter molecule identity/routing checks;
4. align this work with an existing issue such as #114 or #116?

Happy to adapt the scope and implementation style to the maintainers' roadmap.

tdpham2:
Hi @rockyzl,

Thank you for your interest in ChemGraph and for taking the initiative to build these reliability improvements! The features you tested locally, especially the target-scoped routing and deterministic validation, sound like fantastic additions.

To keep the review process manageable, your first suggestion is exactly what we need: please split these into smaller, focused PRs, starting with the smallest validation/routing fixes.

Also, please ensure all PRs are opened against the `dev` branch rather than upstream `main`.

As a quick heads-up, I am currently traveling for a conference, so my reviews may take a bit of time. Looking forward to seeing your first PR!

keceli:
Hi @rockyzl,

[...]

PR #145, opened 2026-07-10
Fix target-scoped single-agent tool validation

rockyzl:
## Summary
- Add a deterministic guard before the single-agent `ToolNode` executes `run_ase`.
- Block follow-up tool calls that reuse a coordinate file from a previous molecule target.
- Feed the agent a `ToolMessage` repair instruction so it can generate/reuse coordinates for the current target instead of running on stale artifacts.
- Cover the stale-target case with routing tests.

## Why
This addresses the smallest validation/routing slice discussed in #135: a follow-up such as `glucose dipolemoment` should not execute `run_ase` on a coordinate file produced earlier for another molecule, such as methane.

## Demo scenario
Run the Streamlit UI:

```bash
streamlit run src/ui/app.py
```

Then use a single chat session:

1. Ask for a property that creates a coordinate artifact for one molecule, for example:

   ```text
   What is the IR spectrum of methane?
   ```

2. Follow up with a different molecule target:

   ```text
   glucose dipolemoment
   ```

3. The routing risk is when the model tries to call `run_ase` with the old file, for example:

   ```json
   {
     "name": "run_ase",
     "args": {
       "params": {
         "input_structure_file": "/tmp/methane.xyz",
         "driver": "dipole"
       }
     }
   }
   ```

Without this guard, that stale tool call can reach `ToolNode`. After this PR, it is blocked and the agent receives feedback to resolve/generate coordinates for the current target, `glucose`, before calling `run_ase`.

```mermaid
sequenceDiagram
    participant User
    participant Agent
    participant Validator
    participant ToolNode

    User->>Agent: What is the IR spectrum of methane?
    Agent->>ToolNode: smiles_to_coordinate_file -> /tmp/methane.xyz
    User->>Agent: glucose dipolemoment
    Agent->>Validator: run_ase(input_structure_file=/tmp/methane.xyz)
    Validator-->>Agent: Blocked: current target is glucose, file belongs to methane
    Agent->>ToolNode: generate/reuse glucose coordinates, then run_ase on glucose
```

The focused test is `test_route_tools_blocks_stale_target_coordinate_reuse` in `tests/test_single_agent_routing.py`.

## Validation
- `git diff --check`
- `PYTHONPATH=src python -m py_compile src/chemgraph/utils/tool_validation.py src/chemgraph/graphs/single_agent.py`
- `PYTHONPATH=src python -m pytest tests/test_single_agent_routing.py -q` -> 6 passed
- `PYTHONPATH=src python -m pytest tests/test_single_agent_routing.py tests/test_graphs.py tests/test_llm_agent.py tests/test_ui_main_interface.py -q` -> 37 passed
- `PYTHONPATH=src python -m pytest -q` -> 168 passed, 6 skipped

Refs #135

### DIFF
--- a/src/chemgraph/graphs/single_agent.py
+++ b/src/chemgraph/graphs/single_agent.py
@@ -21,6 +21,12 @@ from chemgraph.prompt.single_agent_prompt import (
 )
 from chemgraph.utils.logging_config import setup_logger
 from chemgraph.utils.parsing import parse_response_formatter
+from chemgraph.utils.tool_validation import (
+    build_tool_validation_messages,
+    latest_user_query,
+    message_tool_calls,
+    validate_target_scoped_tool_calls,
+)
 from chemgraph.state.state import State
 
 logger = setup_logger(__name__)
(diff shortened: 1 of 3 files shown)


rockyzl:
Hi @keceli and @tdpham2 — when you have a chance, could you please review this focused PR? It is ready and mergeable. The GitHub Actions runs from my fork are currently awaiting maintainer approval. Thank you!

keceli:
Thank you @rockyzl for the PR. This is a tricky failure mode, and it is great that you went for a deterministic guardrail with real routing tests. The graph wiring reads cleanly.

My one bigger-picture worry is the design: the decision rests on re-parsing the free-text query, guessing the molecule from the filename, and string-matching the two. Even once polished, that base tends to drift, e.g. "dipole moment of the water molecule" gives target "water molecule" ≠ species "water", and the default molecule.xyz resolves to a generic "molecule". Each miss usually becomes another special case on the same guessing layer. The sturdier version is probably to track the active target molecule as explicit graph state (set when a molecule is resolved) and give coordinate files provenance, so run_ase checks identity structurally rather than from strings.

One thing I'd fix regardless: when a blocked run_ase shares a message with other parallel tool calls, only the blocked id gets a ToolMessage, so the siblings go unanswered and the next turn can "400 Bad Request". Worth handling all tool_call_ids.

PR #159, opened 2026-07-27
Docs & Fix: connecting_to_argo — compute-node → argo-shim → Argo recipe

JinchuLi2002:
## Summary

Adds `examples/connecting_to_argo/` — README + runnable `test_run.py`
covering both argo-shim topologies (login-node relay, laptop relay)
for reaching Argo from an ALCF compute node.

Also extends `ARGO_LOCAL_OPENAI_MODEL_MAP` in
`chemgraph/models/openai.py` with entries for the GPT-* family.
Previously only `gpt-5.4` was mapped, so other models 400'd with
`Invalid model` because the shim expects display-case (`GPT-4.1-mini`)
but ChemGraph was sending lowercase (`gpt-4.1-mini`) after prefix
stripping.

## Type of change

- [x] Bug fix
- [x] Docs

## How was this tested?

`python examples/connecting_to_argo/test_run.py` on an Aurora compute
node against Argo — LLM answers the "geometry optimization for water
using TBLite" prompt through the two-hop shim (compute → login-node
tunnel → CELS → Argo).

## Checklist

- [x] Branched off `main` and targets `main`
- [x] Single logical change (docs example + one-file bug fix in the
      map that the docs reference)
- [ ] `ruff check .` 
- [ ] `pytest tests/` — not run (no test changes)
- [x] Docs added for the new example

PR #160, opened 2026-07-29
academy: fresh-install fixes

JinchuLi2002:
## Summary

Five small fixes so a fresh clone + `pip install -e .` + `swarm dashboard`
runs the mof-crux-aurora-mock demo end-to-end without any ad-hoc edits.

1. `academy-py` → top-level `dependencies` (was in `[academy]` extra,
   fresh installs missed it; daemon failed at `from academy.handle
   import Handle`). Redis stays optional.
2. Clear `pythonpath_entries` in both site profiles. Old default shadowed
   pip-installed code with a stale sibling checkout; symptom was
   `--exchange-type: invalid choice: 'http'` from an older daemon.py.
3. `mof-crux-aurora` campaign: rename `swarm.tools.pacmof2_mcp` and
   `swarm.tools.globus_flex_mcp` to `chemgraph.academy.tools.*`
   (pre-rename paths that no longer exist).
4. Extend `ARGO_LOCAL_OPENAI_MODEL_MAP` to the full GPT-* family
   (was only `gpt-5.4`; other models 400'd with "Invalid model"
   against local argo-shim).
5. `resolve_campaign` now warns when a user-copy is silently overriding
   a newer shipped example. Doesn't change behavior — makes the
   mismatch visible in the log so operators aren't debugging phantom
   stale paths.

## Not addressed

- `mcp_servers[].command` in both campaign JSONs still hardcodes venv
  paths. Fresh operators still have to edit these; the resolver just
  logs a warning if their edits are being silently overridden by an
  old user-copy (fix 5). Real fix needs env-var expansion in the
  campaign loader — separate design discussion.

## Type of change

- [x] Bug fix
- [x] Docs
- [x] Chore / config cleanup

## Test plan

- [x] Fresh clone → `pip install -e .` on Aurora + Crux; both venvs
      import `chemgraph.academy.tools.mock_mcp` and `academy.handle`.
- [x] `swarm dashboard ... --system crux,aurora` on dev launches both
      agents; mock demo completes end-to-end.
- [x] `examples/connecting_to_argo/test_run.py` on an Aurora compute
      node returns a real ChemGraph response through argo-shim.


pyproject.toml
tdpham2, line 1: Please revert back the changes and keep academy-py optional.

src/chemgraph/models/openai.py
tdpham2, line 1: Can you check if this can be resolved without this change by setting 

export CHEMGRAPH_ARGO_MODEL_FORMAT="wire" # or "argo"

examples/campaigns/mof-crux-aurora/campaign.jsonc
tdpham2, line 1: Can we change all the commands from your specific path to a /path/to/python path? E.g user will provide the python path/venv in a config file.

pyproject.toml
JinchuLi2002: Done in a52b8da — moved academy-py back into the [academy] extra

src/chemgraph/models/openai.py
JinchuLi2002: Yes, this variables resolves it. So the extra mapping is not needed (removed in c24bbad)

examples/campaigns/mof-crux-aurora/campaign.jsonc
JinchuLi2002: Done in a52b8da

src/chemgraph/models/openai.py
tdpham2, line 1: Is this to fix a certain error? If it's simply syntax change/code simplification, please open a separate and focused PR against `main`, and revert this file back.

src/chemgraph/academy/runtime/academy_patch.py
tdpham2, line 1: Not related to close this PR, but can you also report this issue on the Academy repo? I'm sure they will want to know about this issue. 

For us, once this issue is resolved on Academy side, we can remove this patch file.

src/chemgraph/academy/runtime/daemon.py
```diff
@@ -8,11 +8,14 @@
 
 
 import chemgraph.academy.runtime.torch_patch  # noqa: F401  -- patches torch.load if torch present
```
tdpham2: I missed the `torch_patch` in the previous PR. Could you clarify why it is needed?

For future work, not in this PR, please avoid patching dependencies and look for a more sustainable solution. You should at least document the reason for the patch and its intended scope. Once the upstream package fixes the underlying issue, these patches can become stale and difficult to maintain.

examples/connecting_to_argo/README.md
tdpham2, line 1: For the examples, please remove connecting_to_argo/ from this PR and open a PR directly against `main`. Let's keep this PR and `dev` focus on getting the workflow to run.

tdpham2:
Please make sure to update the PR text once the comments above are addressed (e.g academy in pyproject.toml)

examples/campaigns/mof-crux-aurora/campaign.jsonc
```diff
@@ -52,24 +48,28 @@
   "mcp_servers": [
     {
       "name": "mofforge",
-      "command": "/lus/eagle/projects/ChemGraph/jinchu/venvs/academy-swarm-crux/bin/python -m mofforge.mcp.server"
+      "command": "$VIRTUAL_ENV/bin/python -m mofforge.mcp.server"
```
tdpham2, line 74: How to separate Aurora MCP vs Crux MCP? I don't see an Aurora mcp_servers in the config

src/chemgraph/models/openai.py
JinchuLi2002: This change is needed to fix broken references. This is a follow-up step of using `export CHEMGRAPH_ARGO_MODEL_FORMAT="wire" # or "argo"`.

src/chemgraph/academy/runtime/academy_patch.py
JinchuLi2002: Yes, it seems to be an known issue from academy side (https://github.com/academy-agents/academy/issues/421)

src/chemgraph/academy/runtime/daemon.py
JinchuLi2002: The `torch_patch` is  because PyTorch 2.6 flipped torch.load's weights_only default to True as a security default, which breaks loading pickled MLIP checkpoints that fairchem-core (UMA) and mace-torch still ship: `UnpicklingError: Weights only load failed`. 

Expanded the docstring in b44cd5a

examples/connecting_to_argo/README.md
JinchuLi2002: Sounds good, they are removed in 4695f0f, I have reopened PR #159 which is against `main` which focuses on `connecting_to_argo`

examples/campaigns/mof-crux-aurora/campaign.jsonc
JinchuLi2002: The mcp_servers block is a library of MCP definitions shared across sites, not a per-site inventory. Each site's daemon reads the whole list but only spawns the subset the running agent whitelists (see daemon.py:72-77):

```
supervisor = MCPServerSupervisor(
    specs=[spec for spec in campaign.mcp_servers
           if spec.name in agent_spec.mcp_servers],
    ...
)
```
For mof-crux-aurora that resolves to:

Crux (`planner_executor.mcp_servers = ["mofforge", "chemgraph", "pacmof2", "globus"]`) — spawns those four, never touches graspa.
Aurora (`aurora_sim.mcp_servers = ["graspa", "globus"]`) — spawns those two, never touches mofforge/chemgraph/pacmof2.

Nothing idles, if a server isn't in the running agent's whitelist, it isn't started on that site. The site→MCP mapping is implicit through which agent runs where, driven by the `--agents` slice at launch time (federated spawn-site flow, daemon.py:60-61).

The $VIRTUAL_ENV/bin/python in each command (added in 17eb519) is what lets one shared library entry work on both sites: it expands at spawn time from the daemon's own sys.executable, so globus lands on /eagle/.../academy-dev/bin/python on Crux and /flare/.../academy-dev/bin/python on Aurora from the same JSON.

Updated docs to make this clearer in 0abd425

src/chemgraph/academy/runtime/academy_patch.py
tdpham2: Can you create an issue for this (on our repo) and add a link to that thread (on Academy)? So we can keep track of this later. Also make sure to highlight your fix in this PR in the issue.

src/chemgraph/academy/runtime/daemon.py
tdpham2: We don't use PyTorch 2.6 though, right? I think UMA requires PyTorch >= 2.8. Not sure about MACE.

On Aurora, the PyTorch from module load frameworks is 2.10.

examples/campaigns/mof-crux-aurora/campaign.jsonc
tdpham2: Having this make sense: $VIRTUAL_ENV/bin/python. 

Why did the old previous version work though because it only points to python on Crux? Is there any stale code that we should address later?

JinchuLi2002: Great question. Previously since we've been only working with the mock campaign, before the `$VIRTUAL_ENV` patch, there are hardcoded path for each mcp server (https://github.com/argonne-lcf/ChemGraph/blob/dev/examples/campaigns/mof-crux-aurora-mock/campaign.jsonc).

Note for stale code: There's a helper in `mcp_supervisor.py` that wraps each subprocess with a` python -c "import chemgraph.academy.runtime.torch_patch; ..." `preamble. That works fine right now because every MCP uses `$VIRTUAL_ENV/bin/python` = the daemon's own venv, which has `chemgraph` installed. But if someone ever adds an MCP command that points at a totally different Python (some external tool's venv), the import chemgraph... in the preamble would fail.

src/chemgraph/academy/runtime/daemon.py
JinchuLi2002: I see. Dropped torch_patch in [671f030](https://github.com/argonne-lcf/ChemGraph/pull/160/commits/671f030677841e572cab0843aea8c42120ba14fa). I'm currently trying to rerunning E2E  but the queue time is a bit long.

src/chemgraph/academy/runtime/academy_patch.py
JinchuLi2002: created #161

PR #162, opened 2026-07-31
Add VASP and Quantum ESPRESSO plane-wave DFT calculators

reowszer:
## Summary

Adds two subprocess-based DFT calculators to ChemGraph, VASP and Quantum ESPRESSO, so the agent can run plane-wave DFT through the same `run_ase` path already used for EMT, MACE, and other calculators. These are the primary target engines for long-running calculations.

The calculators are periodicity-aware, which is the core correctness feature of this PR:

* **PBC-aware k-point mesh.** A non-periodic molecule, with all `pbc` values set to False, no longer receives a spurious Monkhorst-Pack mesh. QE emits `K_POINTS gamma`; VASP pins the mesh to `[1, 1, 1]`, enables `gamma`, and removes `kspacing`. Otherwise, a stray `KSPACING` entry in the INCAR could override the gamma point specified by KPOINTS. Periodic systems retain their configured meshes. Constructing a calculator without atoms also preserves the configured mesh for backward compatibility. Slabs and wires use a per-axis mesh: periodic axes retain their configured subdivisions, while non-periodic vacuum axes are forced to a single point.
* **Non-periodic cell handling.** A molecule without a cell is centered inside a finite vacuum box, controlled by the `vacuum` field, before it reaches the input writer. Plane-wave QE and VASP calculations cannot run with a zero cell. The molecule's `pbc` values are still False when the calculator is built, so it follows the gamma-point branch above.
* **QE command sanitization.** A legacy full `ASE_ESPRESSO_COMMAND`, such as `mpirun -np 4 pw.x -in PREFIX.pwi > PREFIX.pwo`, is reduced to only the launcher and executable before being passed to `EspressoProfile`. This prevents ASE's own `-in` handling from being duplicated and ensures that no `>` token leaks into the argument vector. Runtime flags such as `-npool` and `-ndiag` are preserved.
* **VASP INCAR escape hatch.** An `input_data` field passes advanced INCAR parameters, including ISMEAR, SIGMA, ALGO, LREAL, MAGMOM, and others, through to ASE. Keys are converted to lowercase during the merge, with documented last-value-wins precedence. VASP also gains a `gamma` field. Molecular calculations default to Gaussian smearing with `ismear=0`, which is appropriate for isolated systems.
* **Fully periodic cells for VASP.** ASE refuses any structure whose `pbc` is not all True (`check_pbc`) before writing a single input file, so a centered molecule, a slab `[T,T,F]`, and a wire `[T,F,F]` all failed setup and only a bulk solid reached `vasp_std`. The cell is now marked fully periodic for VASP, after the calculator is built so the k-mesh above still follows the input periodicity. A plane-wave code has no aperiodic mode, so this states what VASP was always going to do; the vacuum is what keeps the residual image interaction small. No dipole or monopole correction is applied, so a polar or charged species needs a larger box or an explicit `IDIPOL` / `LDIPOL` through `input_data`. QE keeps its `pbc` untouched: `pw.x` has no equivalent check and `write_espresso_in` ignores `pbc`, so the flip would only leak a spurious periodic flag into QE results.

Because all existing calculators expose a zero-argument `get_calculator()` method, `ase_core.py` passes `atoms` only to `EspressoCalc` and `VaspCalc`, selected through an `isinstance` check.

Gas-phase thermochemistry keys off the input's periodicity, recorded before that flip. A molecule this code boxed stays eligible for `IdealGasThermo`; a structure that arrived periodic keeps its periodicity and is still refused, so a crystal never collects rigid-rotor statistics.

### Plane-wave tools: `run_qe` and `run_vasp`

Two calculator-pinned tools plus a runnable example.

* **Thin wrappers.** `run_qe` and `run_vasp` are `@tool` wrappers that build an `ASEInputSchema` with a fixed calculator and call `run_ase_core`; dispatch already routes the `espresso` and `vasp` aliases, so the core is reused directly. `QEInputSchema` / `VaspInputSchema` (new `schemas/plane_wave_input.py`) subclass `ASEInputSchema` and pin the calculator field, so the model sees a focused single-engine tool and inherits the common fields for free.
* **Registration.** `single_agent.py` adds each tool to the default set gated on engine availability; `tool_call_eval.py` extends the `run_ase` special-case to both for schema-aware argument comparison.
* **Token cost.** The two schemas add ~3248 tokens (tiktoken `cl100k_base`: `run_qe` 1621 + `run_vasp` 1627), about half of `run_ase` (3184) each because the calculator type is fixed.

The example in `examples/plane_wave_tools/` binds the tools to a `single_agent` ChemGraph over Argo and runs a bulk-Si relaxation prompt, with a `README.md` covering the full setup (`ASE_ESPRESSO_COMMAND`, `ESPRESSO_PSEUDO`, the VASP equivalents, and the Argo tunnel).

The example writes the Si cell itself, from Materials Project `mp-149` when `MP_API_KEY` is set and from `ase.build.bulk` otherwise, and names that file in the prompt. No tool in the `single_agent` set can build a crystal: the only structure source is `smiles_to_coordinate_file`, which is RDKit-backed and returns an isolated molecule with no cell, so a bare "bulk silicon" prompt yielded one lone Si atom. The LLM still chooses the tool and fills in the driver, cutoff, k-mesh, and pseudopotentials; only the lattice is supplied. A crystal-building tool would remove even that, and belongs in its own PR.

### Part of a 3-PR series

This is the first of three PRs that together add long-running-calculation support for subprocess DFT. They are intended to land in the following order:

1. **`feature/dft-calculators` (this PR):** adds the VASP and Quantum ESPRESSO calculators.
2. `feature/longrun-cap-resume`: adds the wall-clock cap, run manifest, and `chemgraph resume`. It overlaps with this PR at the file level but is functionally independent, and the branches have been verified to merge cleanly.
3. `feature/aurora-qe-example`: adds a long-running cap-to-resume example on Aurora that validates the full workflow using real QE DFT. It depends on both earlier PRs.

The example PR will remain in Draft until the calculators and cap PRs merge.

## Related issues

None

## Type of change

* [ ] Bug fix
* [x] New feature
* [ ] Docs
* [ ] Chore / refactor / CI

## How was this tested?

* Hermetic unit tests in `tests/test_calculators.py` assert against both the parameter dictionaries of the constructed ASE calculators and the actual input files written to disk. These checks cover QE's `K_POINTS gamma` and finite `CELL_PARAMETERS`, as well as VASP's `kpts`, `gamma`, and INCAR parameters, and add dispatch-parity tests proving `run_qe` / `run_vasp` drive the same `run_ase_core` path as `run_ase`. The tests therefore validate the generated inputs, not only the schema. No DFT calculations run in CI.
* The QE path is validated against real `pw.x` DFT on ALCF Aurora. The full agent stack was run with a prompt that named Quantum ESPRESSO but named no tool; the LLM autonomously selected `run_qe`, and real `pw.x` relaxed bulk Si with BFGS to convergence at approximately -308.189 eV. This was confirmed on both LLM backends independently: the ALCF inference endpoint (gpt-oss-120b) and Argo (gpt-4.1-mini).
* The VASP path is validated against real `vasp_std` (VASP 6.5.1, 8 MPI ranks, PBE PAW) on a VASP-licensed cluster reached through Argo, with the agent again choosing `run_vasp` unprompted. ENCUT follows one rule throughout: 1.3 times the hardest element's POTCAR ENMAX, rounded up to the nearest 25. Energies in eV: an H2O molecule at -14.21983871 on a 1x1x1 gamma mesh with `ISMEAR=0`, an Al slab at -10.65570601 on a 4x4x1 mesh with the vacuum axis correctly collapsed, and bulk Si at -10.83792607, matching a direct ASE run of the same cell. The molecule and the slab were hard failures before the periodicity fix above. Running the example end to end gave -10.83792607 eV from the `ase.build` cell and -10.84103139 eV from `mp-149`; a lattice scan at these settings puts the minimum near a=5.47, so the 3 meV gap reflects `mp-149` sitting closer to it.
* Both real-engine runs exercise the code path and flow on real hardware. Neither is a physical-accuracy benchmark.
* `ruff check .` passes.
* `pytest tests/ -k "not tblite"` passes with 374 passed, 28 skipped, and 2 deselected. Each new periodicity and thermochemistry test was confirmed to fail with its fix reverted.

## Checklist

* [x] Branched off the latest `main` and targets `main`
* [x] PR is focused on a single logical change and was split where necessary
* [x] `ruff check .` passes
* [x] `pytest tests/ -k "not tblite"` passes, including additional tests if Academy or backend code was touched
* [x] Added or updated tests for the change
* [x] Updated docs or README for any user-facing change

### DIFF
--- a/src/chemgraph/graphs/single_agent.py
+++ b/src/chemgraph/graphs/single_agent.py
@@ -7,8 +7,11 @@ from langgraph.checkpoint.memory import MemorySaver
 from langgraph.prebuilt import ToolNode
 from chemgraph.tools.ase_tools import (
     run_ase,
+    run_qe,
+    run_vasp,
     extract_output_json,
 )
+from chemgraph.schemas.ase_input import get_available_calculator_names
 from chemgraph.tools.cheminformatics_tools import (
     molecule_name_to_smiles,
     smiles_to_coordinate_file,
@@ -25,6 +28,30 @@ from chemgraph.utils.parsing import parse_response_formatter
 from chemgraph.state.state import State
 
 logger = setup_logger(__name__)
+
+
+def _plane_wave_tools() -> list:
+    """Return the plane-wave DFT tools whose engine is detected on this host.
+
+    ``run_qe`` / ``run_vasp`` are calculator-pinned wrappers over ``run_ase``.
+    Each is only useful when its engine (pw.x / VASP binary + pseudopotentials)
+    is installed, so we offer it only when the corresponding calculator survived
+    the availability gate in ``chemgraph.schemas.ase_input`` -- mirroring how the
+    ``CalculatorUnion`` itself is filtered. This keeps them out of the prompt on
+    hosts that cannot run them (and off the token budget there).
+
+    Returns
+    -------
+    list
+        The subset of ``[run_qe, run_vasp]`` whose engine is available.
+    """
+    available = set(get_available_calculator_names())
+    tools = []
+    if "EspressoCalc" in available:
+        tools.append(run_qe)
+    if "VaspCalc" in available:
+        tools.append(run_vasp)
+    return tools
 
 
 def _tool_call_signature(tool_calls) -> tuple:
@@ -307,6 +334,7 @@ def ChemGraphAgent(
             extract_output_json,
             calculator,
         ]
+        tools.extend(_plane_wave_tools())
         if human_supervised:
             tools.append(ask_human)
     elif human_supervised and ask_human not in tools:
@@ -499,6 +527,7 @@ def construct_single_agent_graph(
                 extract_output_json,
                 calculator,
             ]
+            tools.extend(_plane_wave_tools())
             if human_supervised:
                 tools.append(ask_human)
         elif human_supervised and ask_human not in tools:
--- a/src/chemgraph/schemas/plane_wave_input.py
+++ b/src/chemgraph/schemas/plane_wave_input.py
@@ -1,0 +1,121 @@
+"""Calculator-pinned input schemas for the plane-wave DFT tools.
+
+``run_ase`` accepts any installed calculator through the ``CalculatorUnion`` on
+:class:`~chemgraph.schemas.ase_input.ASEInputSchema`, so its whole tool schema
+(every calculator's fields) is serialized into every LLM request. The dedicated
+``run_qe`` / ``run_vasp`` tools instead each pin a single engine, so the model
+sees only that engine's parameters. These subclasses reuse every common ASE
+input field (``input_structure_file``, ``driver``, ``optimizer``, ``fmax``,
+``steps`` ...) from :class:`ASEInputSchema` and override only ``calculator`` to
+the concrete plane-wave type.
+
+The parent's ``_validate_calculator_type`` coerces + availability-gates the
+calculator against ``available_calculator_classes`` (engines detected at import
+time). That gate is right for ``run_ase`` (the union only offers installed
+engines) but wrong here: these schemas are also constructed in hermetic tests
+and by ``tool_call_eval`` on machines with no pw.x / VASP binary. So each
+subclass replaces the validator with a narrow coercion that instantiates its own
+calculator type WITHOUT the availability probe. Registration of the *tools*
+themselves still gates on availability (see ``graphs/single_agent.py``), so an
+engine that is not installed is never offered to the model on a real run.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from pydantic import Field, model_validator
+
+from chemgraph.schemas.ase_input import ASEInputSchema
+from chemgraph.schemas.calculators.espresso_calc import EspressoCalc
+from chemgraph.schemas.calculators.vasp_calc import VaspCalc
+
+
+def _coerce_pinned_calculator(data: Any, calc_cls: type, canonical_type: str) -> Any:
+    """Coerce ``data['calculator']`` into ``calc_cls`` without the availability gate.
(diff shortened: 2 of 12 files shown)


src/chemgraph/graphs/single_agent.py
tdpham2, line 1: For this, binding the tools to the main graph (single_agent) will give issues with evaluation - meaning we'd have to rerun the evaluation first to make sure this does not negatively affect the agent's performance (especially for smaller models).

Instead of this, for now, can you follow the example in https://github.com/argonne-lcf/ChemGraph/pull/171? So create a separate graph under `graph`, bind run_qe and run_vasp, and other tools that you see needed for a plane-wave calculation workflow. Tools like fetch from material project (as you added in the examples/), and remove tools that are not needed.

src/chemgraph/schemas/plane_wave_input.py
tdpham2, line 1: Could you explain why this is needed?

I thought there is already a check under ase_input.py. If there isn't a binary/variable, the schemas of those calculators (VASP/QE) is not part of the tool schema.

src/chemgraph/utils/tool_call_eval.py
tdpham2, line 1: I kind of understand this. However, should we start building something like

`ASEPBCInputSchema` instead of one schema for each calculator? Particularly handle all PBC calculations for QE, VASP and potentially CP2K?

tests/test_calculators.py
tdpham2, line 1: Can you consolidate the tests? All tests are hermetic, so only keep the important ones.

PR #163, opened 2026-07-31
Add wall-clock cap, run manifest, and chemgraph resume for long-running calculations

reowszer:
## Summary

Adds ChemGraph's long-running-calculation feature: a soft wall-clock cap that
stops an ASE optimization at a step boundary, a durable run manifest that records
what finished, and a `chemgraph resume` path that continues a capped run in a
later allocation without recomputing. This lets a calculation that outlasts a
single scheduler allocation finish across several of them.

The four commits are each one logical layer:

- **Wall-clock cap (`ase_core.py`).** The optimizer yields at each step boundary;
  the deadline (`CHEMGRAPH_ALLOCATION_DEADLINE`/`_SECONDS` minus a safety
  `_MARGIN`) is checked at each yield. When a step completes within the budget,
  a restart file exists and the run resumes from it; a cap that fires before
  step 1 leaves no restart and is reported as such. The cap is soft: an in-flight
  SCF is left untouched, so the margin must exceed one step's wall time.
- **Run manifest (`memory/manifest.py`).** A durable per-session JSON record of
  each executed cost-bearing tool call (name, args, result-file path, wall_time)
  and the pending next step, written incrementally with an atomic tmp-plus-replace.
  It is independent of the in-memory checkpointer and the turn-end SessionStore, so
  a completed step survives a walltime kill. Reads are defensive: a wrong-shape
  file falls back to a fresh default, and `render_for_context` tolerates malformed
  steps, because it runs unguarded on the resume path.
- **CLI (`cli/jobs.py`, `cli/commands.py`, `cli/main.py`).** `chemgraph jobs
  {list,status,results}` inspects the persisted HPC job batches offline, and
  `chemgraph resume` continues a capped session.
- **gitignore.** Ignores the per-run cap and manifest artifacts.

### When this helps

Even though starting the next allocation is still manual today, the value is
concrete: the cap preserves completed computational work that would otherwise be
discarded.

- **A single optimization that needs more than one walltime.** A geometry
  optimization of a mid-size system can run many optimizer steps at minutes to
  hours each, more than a single queue slot grants. Without the cap, the scheduler
  SIGKILLs the run partway through a step and the completed optimization progress
  is not durably handed off to the next allocation, so the next submission starts
  from zero and the run never finishes. With the cap, the run stops at a step
  boundary and saves the partial geometry; a later `chemgraph resume` continues
  from the last completed step, so the work
  finishes across a few allocations with no step recomputed.
- **Many short jobs schedule better than one long allocation.** Splitting a long
  calculation into several standard-length jobs is friendlier to the queue: short
  jobs start sooner and backfill more easily than a single special long-walltime
  allocation that may wait a long time or be denied.
- **Insurance against an underestimated walltime.** Even a run expected to fit can
  overrun when convergence is slower than planned. Without the cap that is a
  SIGKILL with total loss; with the cap the worst case is stopping at the last
  completed step, which is always resumable. This safety holds with zero
  automation layered on top.

### Part of a 3-PR series

This is the second of three PRs that together add long-running-calculation support
for subprocess DFT. They are meant to land in order:

1. feature/dft-calculators: the VASP and Quantum ESPRESSO calculators.
2. **feature/longrun-cap-resume (this PR):** the wall-clock cap, run manifest, and
   `chemgraph resume`. It overlaps with the calculators PR at the file level but
   is functionally independent, and the branches have been verified to merge
   cleanly.
3. feature/aurora-qe-example: a runnable Aurora example that validates the seam on
   real QE DFT. Depends on both earlier PRs.

The example PR opens as Draft until this one and the calculators PR merge.

## Related issues

None

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Chore / refactor / CI

## How was this tested?

- New unit tests cover each layer on the in-process EMT/MACE calculators, with no
  HPC or DFT needed: `test_cap.py`, `test_allocation_cap.py`, `test_mace_cap.py`
  (the cap fires at a step boundary and leaves a resumable restart file),
  `test_manifest.py` and `test_job_tracker_hardening.py` (atomic write,
  wrong-shape and malformed-step tolerance), `test_cap_manifest_integration.py`
  and `test_resume_injection.py` (cap to manifest to resume end to end),
  `test_cli_jobs.py`, `test_cli_resume.py`, `test_cli_cap_handoff.py`.
- The same seam is validated against real `pw.x` DFT on ALCF Aurora in the
  companion example PR (bulk Si and a centered H2O molecule, calc layer and full
  agent stack).
- `ruff check .` passes. `pytest tests/ -k "not tblite"` passes (371 passed, 28
  skipped, 2 deselected).

## Checklist

- [x] `ruff check .` passes

The `chemgraph jobs` and `chemgraph resume` subcommands are self-documenting through
`--help`, and the end-to-end workflow documentation is provided in the companion
example PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,7 @@ vib*.traj
 
 # Kubernetes secrets (keep secrets.yaml.template, ignore actual secrets)
 k8s/secrets.yaml
+
+# Long-run cap / manifest run artifacts (session log dirs, test outputs)
+cg_logs/
+tests/*_output.json
--- a/src/chemgraph/agent/llm_agent.py
+++ b/src/chemgraph/agent/llm_agent.py
@@ -1,7 +1,9 @@
 import asyncio
 import datetime
+import json
 import os
 import time
+from pathlib import Path
 from typing import Callable, Collection, List, Optional
 import uuid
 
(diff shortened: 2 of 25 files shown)


tdpham2:
@reowszer Thanks for the summary. I understand this PR as three layers. Is the following an accurate description?

 1. A soft deadline in run_ase_core.
 2. ASE-level checkpoint artifacts for optimization and vibrational calculations.
 3. An agent-level manifest that is rendered into the prompt for chemgraph resume.

I would avoid the general claim that many short jobs schedule better than one long allocation. Shorter requests may sometimes fit backfill opportunities, but creating many jobs also adds scheduling overhead. Work that fits within an allocation should generally be distributed internally (with workflow manager like Parsl). Aurora’s production queues are also capability oriented, with a current minimum request of 256 nodes.

I think this is the right direction. I think it would be easier to review and extend if separated into smaller PRs:

1. Checkpoint and restart for run_ase. Start with geometry optimization. Maybe write/store ASE trajectory, and provide an explicit way to restart from a selected checkpoint (maybe in the run_ase function). Let's only use the last geometry (instead of wavefunction file) as checkpoint/restart method for now.
2. A calculation database. We already use SQLite for conversation history, so we could either extend that schema or introduce a separate CalculationStore. Each calculation should have a stable ID and record its session ID, tool name, complete parameters, state, results, and checkpoint artifacts. Large outputs and native checkpoint files can remain on the shared filesystem, with the database storing their paths and and maybe checksums.
3. Calculation resume. When agent want to resume (which the CLI already supports the --resume for conversation by appending previous conversation history), attach the database entry (from (2)) with calculations details that agent maybe able to retrieve from the database. Agent can decide to resume via (1). 

Let me know what you think and we can update the plan accordingly.

Issue #146, opened 2026-07-16
[REFACTOR] Rework ChemGraph installation


tdpham2:
### Problem

ChemGraph currently includes MLIP packages (MACE) in `pyproject.toml`. This works for local installations, where `pip` can manage the PyTorch version and related dependencies.

However, this causes problems on HPC systems, particularly Aurora. On Aurora, users must load the system-provided PyTorch environment through:

```bash
module load frameworks
```

This PyTorch build is configured for Intel GPU/XPU support. Installing MACE, FairChem, or ChemGraph through `pip` may replace or conflict with the system-provided PyTorch version. As a result, the MLIP packages may only work on CPUs or may fail due to incompatible dependencies.

This is likely a broader issue across HPC systems, where PyTorch builds are selected and optimized for specific hardware and software environments.

### Proposed direction

Keep the core ChemGraph installation lightweight:

* Remove MACE, FairChem, and other MLIP packages from the required dependencies.
* Retain TBLite/xTB as the default lightweight computational backend for demonstrations and basic testing.
* Provide MLIP integrations as optional dependencies. Local deployment can still install these dependencies.
* Recommend that HPC users configure the appropriate PyTorch environment and install the desired MLIP packages separately before installing ChemGraph.

For example, the recommended installation order on an HPC system could be:

1. Load the system-provided PyTorch environment.
2. Install MACE, FairChem, or other required MLIP packages in a compatible environment.
3. Install ChemGraph without modifying the existing PyTorch installation.

### Possible dependency structure

```toml
[project.optional-dependencies]
mace = [
    "mace-torch",
]

fairchem = [
    "fairchem-core",
]

mlips = [
    "mace-torch",
    "fairchem-core",
]
```

We should also ensure that ChemGraph imports these packages only when the corresponding tools are used and provides a clear error message when an optional dependency is unavailable.

tdpham2:
Current conversation converges to keep using MACE in the default installation, with MACE-polar being our default model. FYI @reowszer

PR #169, opened 2026-08-05
Make mace-torch an optional extra

reowszer:
## Summary

`mace-torch` is currently a required dependency, although the schema path
already treats MACE as optional. `ase_input.py` registers `MaceCalc` only when
`_engine_available("mace")` is true, leaving `pyproject.toml` and several tests
as the remaining gaps.

On a cluster without MACE installed, MACE-specific tests errored during fixture
setup where a skip was the intended outcome. The fixtures construct an
`ASEInputSchema` with `mace_mp`, which is unregistered in that environment and
fails Pydantic validation. The `mace-torch` dependency also complicates UMA
installation, because it requires `e3nn==0.4.4` while the `uma` extra requires
`e3nn>=0.5`.

This PR moves `mace-torch` to a `mace` extra, defers torch imports that were
previously satisfied transitively, and adds the missing test guards.

### User-visible behavior change

After this change, `pip install chemgraph` no longer installs `mace-torch` or
registers `MaceCalc`.

* If `calculator` is omitted, selection proceeds to the first available class
  in the `ase_input` priority list (FAIRChem, MACE, AIMNET2, TBLite, EMT,
  NWChem, Orca). On the tested minimal installation, this resolves to EMT.
* Explicitly requesting `mace_mp` raises a Pydantic `ValidationError` listing
  the available calculators. No fallback calculator is substituted.
* `pip install "chemgraph[mace]"` restores the previous behavior.
* `get_calculator_selection_context()` already injects real availability into
  the prompt, so the agent selects from what is installed.

**Existing environments or CI jobs that install plain `chemgraph` and
explicitly use `mace_mp` will need `chemgraph[mace]` after this change.** This
is the main compatibility tradeoff for maintainers to consider.

A narrower alternative is to keep `mace-torch` required and add only the test
guards. That fixes the fixture errors and leaves the UMA conflict and
installation size unchanged.

### Benefits

* **UMA installation no longer requires manually removing `mace-torch` and
  `e3nn`.** The README walked users through hand-editing `pyproject.toml`,
  installing `[uma]`, then restoring the file, and `dependency_tests.yml` did
  the CI equivalent with `pip uninstall -y mace-torch e3nn`. Neither is needed
  now.
* **The tested core installation resolves 35 fewer packages**, including
  `torch`, `triton`, `nvidia-cublas`, `nvidia-cudnn-cu13`, and other CUDA
  runtime packages. Users running EMT or the subprocess DFT engines no longer
  download any of it.
* **Tests for unavailable engines now skip as intended.**

### Deferred torch imports

`torch` is not a declared ChemGraph dependency and was previously installed
transitively through `mace-torch`. Because `mace_calc.py` and
`fairchem_calc.py` imported torch at module scope, making MACE optional caused
`chemgraph.schemas.ase_input` to fail on a core installation, also blocking EMT
and TBLite. These imports are now deferred to the functions that require them,
following the `AGENTS.md` rule on optional dependencies.

`fairchem_calc.py` read `torch.cuda.is_available()` in a Pydantic field
default, which runs at class definition. The field now defaults to the literal
`"cpu"`, and GPU detection occurs when the calculator is built. This preserves
the `"default"` entry in the generated JSON schema used by `tool_call_eval` and
`bind_tools`. That module already treated `fairchem.core` as optional and
applied no equivalent guard to torch, so this gap predates the PR.

### Deliberate non-changes

* Direct `run_ase_core` calls still bypass the availability gate. Requesting
  `mace_mp` without MACE therefore raises `ModuleNotFoundError`, while the
  schema path raises a validation error listing available calculators. This
  pre-existing inconsistency affects every gated calculator.
* Several defaults, documentation examples, and few-shot prompts still select
  `mace_mp`: `config.toml`, `src/ui/config.py`, the copies of that block in
  `README.md` and `docs/configuration_with_toml.md`, and
  `src/chemgraph/prompt/`. These references are intentionally unchanged here.
* `environment.yml` names `mace-torch==0.3.13` explicitly, so the conda path is
  unaffected. A pip `[mace]` installation remains unpinned and may resolve to a
  different version.
* The UMA CI job manually installs `fairchem-core==2.13.0` while the `uma`
  extra pins `2.3.0`. That mismatch, and the Python 3.13 resolution failure the
  pin causes, are unchanged in this PR.
* No MACE-specific matrix leg is added to `dependency_tests.yml`, because the
  three full-suite jobs already install `.[mace]`.

Handling these as follow-ups would keep this PR focused, though any of them can
be folded in on request.

### Overlap with #162

The only conflict with #162 is the `calculator` field docstring in
`ase_input.py`. #162 adds VASP and Quantum ESPRESSO to the list there, and this
PR corrects the ordering and the "use MACE or FAIRChem" line that a core
install no longer satisfies. Everything else merges cleanly, including
`ase_core.py`. Since #162 is further along, merging it first and rebasing this
branch afterward seems best.

## Related issues

None

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Docs
- [x] Chore / refactor / CI

## How was this tested?

* `pytest tests/ -k "not tblite"` in a core installation with `tblite`,
  `parsl`, and `globus_compute` installed: **305 passed, 29 skipped, 2
  deselected**.
* The same suite with MACE hidden through a patched
  `importlib.util.find_spec`: **296 passed, 38 skipped, 2 deselected**. The
  same run on the parent commit produces **8 fixture-setup errors**, 4 in
  `test_mace.py` and 4 in `test_tools.py`, confirming the original failure
  mode.
* With `torch`, `mace`, and `fairchem` all hidden, `chemgraph`,
  `schemas.ase_input`, `tools.ase_tools`, `agent.llm_agent`, and
  `graphs.single_agent` all import successfully. A package-wide import sweep
  covering `src/ui/` found no other top-level imports of optional engines.
* Dry-run resolution produced **245 packages** for the core install, with
  `torch`, `mace-torch`, and `e3nn` all absent, and **280** for `.[mace]`, with
  `torch 2.13.0`, `mace-torch 0.3.16`, and `e3nn 0.4.4` present. A fresh
  editable `.[mace]` install succeeded and `python -m pip check` passed.
* The UMA dependency job passes without the forced uninstall, including
  `pip check`.
* In a fairchem-present, MACE-absent environment, all three `test_faircalc.py`
  tests errored during fixture setup before the guard and skipped cleanly
  afterward.
* `ruff check .` passes.

No DFT calculations were run; validation covered tests, imports, and dependency
resolution only.

## Checklist

- [x] `ruff check .` passes

### DIFF
--- a/.github/workflows/conda-tests.yml
+++ b/.github/workflows/conda-tests.yml
@@ -63,7 +63,8 @@ jobs:
     - name: Install ChemGraph with MACE support
       shell: bash -l {0}
       run: |
-        pip install --no-cache-dir -e .
+        # MACE is an optional extra, and this job exists to exercise it.
+        pip install --no-cache-dir -e ".[mace]"
         pip install --no-cache-dir pytest
     
     - name: Run tests with MACE
@@ -107,18 +108,10 @@ jobs:
   #     run: |
   #       conda install -c conda-forge nwchem -y
 
-  #   - name: Temporarily modify pyproject.toml for UMA e3nn compatibility
-  #     shell: bash -l {0}
-  #     run: |
-  #       echo "Backing up pyproject.toml to pyproject.toml.original..."
-  #       cp pyproject.toml pyproject.toml.original
-  #       echo "Commenting out mace-torch from pyproject.toml..."
-  #       # This sed command finds lines starting with optional whitespace,
-  #       # then "mace-torch>=0.3.13", and prepends a '#' to the matched line.
-  #       sed -i 's/^[[:space:]]*"mace-torch>=0.3.13",/#&/' pyproject.toml
-  #       echo "pyproject.toml after modification:"
-  #       cat pyproject.toml
-    
+  #   # mace-torch is an optional extra, so a plain install brings in neither
+  #   # it nor the old e3nn, and the pyproject edit this job used to perform
+  #   # is no longer needed.
+
   #   - name: Install ChemGraph with UMA support
   #     shell: bash -l {0}
   #     env:
@@ -145,14 +138,4 @@ jobs:
   #     run: |
   #       python -m pytest tests/ -v 
     
-  #   - name: Restore pyproject.toml
-  #     shell: bash -l {0}
-  #     if: always() # Ensures this step runs even if previous steps fail
-  #     run: |
-  #       echo "Restoring pyproject.toml from pyproject.toml.original..."
-  #       if [ -f pyproject.toml.original ]; then
-  #         mv pyproject.toml.original pyproject.toml
-  #         echo "pyproject.toml restored."
-  #       else
-  #         echo "Backup pyproject.toml.original not found. Cannot restore."
-  #       fi 
+ 
--- a/.github/workflows/dependency_tests.yml
+++ b/.github/workflows/dependency_tests.yml
@@ -28,30 +28,22 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         if [ "${{ matrix.install_extras }}" == "uma" ]; then
-          # 1. Install the main package (installs mace-torch + old e3nn)
-          pip install .
-          
-          # 2. FORCE UNINSTALL the conflicting packages
-          echo "💥 Forcefully removing mace-torch and e3nn for UMA compatibility..."
-          pip uninstall -y mace-torch e3nn
-          
-          # 3. MANUALLY install the UMA dependencies.
+          # mace-torch is an optional extra now, so a core install brings in
+          # neither it nor the old e3nn and there is nothing to remove first.
           # torchtnt 0.2.4 (via fairchem-core) imports pkg_resources, which was
-          # removed in setuptools 82.
-          # (We cannot use pip install .[uma] because it would trigger the resolution error)
+          # removed in setuptools 82, so setuptools stays pinned below that.
+          pip install "setuptools<82" .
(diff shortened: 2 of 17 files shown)


tdpham2:
Tagging @keceli here. I think this is more of design choice and user-facing question, as I also made an issue earlier in #146.

If we keep MACE in core dependencies:
- Single-step installation. User runs `pip install chemgraphagent` and can quickly run MACE with ChemGraph, which is open-source. Faster to demo. For UMA, user needs to go to HuggingFace, sign agreement before accessing the model. 

If we remove MACE from core dependency:
- Multiple-step installation. User needs to run `pip install chemgraphagent`, and maybe followed by `pip install chemgraphagent[mace]`.
- Lighter dependency as noted above - quicker installation.
- Current `pip install mace` doesn't work on Aurora XPU. If user wants to run MACE on XPU, they still need to make their own fork and build their environment there. Bundling MACE in ChemGraph thus becomes unnecessary for some HPC deployment.

tdpham2:
@keceli Do you have any thoughts on this PR? We discussed to keep use MACE POLAR as our default model. Should we still keep MACE in our core dependency? Unless MACE is updated in future release, `pip install mace` doesn't work out of the box on Aurora XPU.

tdpham2:
Closing this since we pin mace-torch == 0.3.16 for mace polar.

PR #171, opened 2026-08-07
Make run_docking a default single-agent tool (CLI/Streamlit discovera…

btoheeb1:
…ble)

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Chore / refactor / CI

## Checklist

- [ ] Branched off the latest `main` and targets `main`
- [ ] PR is focused on a single logical change (split if it grew large)
- [ ] `ruff check .` passes
- [ ] Added/updated tests for the change

### DIFF
--- a/src/chemgraph/graphs/single_agent.py
+++ b/src/chemgraph/graphs/single_agent.py
@@ -15,7 +15,6 @@ from chemgraph.tools.cheminformatics_tools import (
 )
 from chemgraph.tools.report_tools import generate_html
 from chemgraph.tools.generic_tools import calculator, ask_human
-from chemgraph.tools.docking_tools import run_docking
 from chemgraph.prompt.single_agent_prompt import (
     single_agent_prompt,
     formatter_prompt,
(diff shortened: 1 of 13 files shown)


src/chemgraph/graphs/single_agent.py
tdpham2, line 1: As I mentioned in the chat, let's not include the `run_docking` tool in ChemGraph's default tool. In the future I may have a fix to include as default tool, but as of now, adding this will affect our evaluation leaderboard.

Instead could you:

1. Add a separate graph under: `graphs/` (e.g docking_graph). Please use a name that best describe how you see this graph will grow in terms of new tools in the future. Use the same setup as single_agent graph, and only modify the default tool list (add run_docking as default tool, and/or modify the list of default tools). Only include tools in this graph that you see are helpful for users when they study this topic. If none is helpful, simply remove them.
2. In `prompts`, add a new prompt file with the same name as the graph added in (1). Make sure the graph in (1) use the new prompt file.
3. In `llm_agent.py`, add a workflow_type for graph in (1).
4. Modify the example in `examples/` if needed.
5. Test with (a) CL, (b) UI and (c) the example

PR #173, opened 2026-08-12
feat: single_agent_iri workflow for ALCF IRI Facility API

JinchuLi2002:
## Summary

Adds a new `single_agent_iri` workflow that exposes ALCF's IRI Facility API (https://api.alcf.anl.gov, 43 endpoints) to the LLM as **seven category-dispatcher tools** with a discovery protocol.

Rather than binding one `@tool` per endpoint (~8k tokens of schemas upfront), each category tool accepts an `action` enum and uses `list_actions` / `describe` for on-demand schema discovery (~1k tokens upfront). Trades one extra LLM turn on cold calls for a ~8x reduction in the tool-schema footprint.

### Discovery protocol

- `What tool categories do you have?`
- `List every action the alcf_filesystem tool supports, and describe the schema of the `ls` action in detail.`

### One hop

- `Which ALCF machines are currently up, and which are down or in unknown state?`

### Two hop

- `Find the ChemGraph project and list all its allocations across every machine, including how much has been used.`

### Three hop

- `Health check: for every compute machine that's currently up, tell me whether my project has allocation left and how many jobs I have in the queue right now.`
- `Would a Crux job survive right now? Check Crux is up, that ChemGraph has positive allocation on Crux, and if either check fails suggest which other ALCF machine I should try instead.`

### Compute — real PBS actions

- `Show the status of Crux job 246115 and describe what happened.`
- `Submit a smoke-test job on Crux: run /bin/bash -lc 'echo hello; sleep 10' under the ChemGraph account, 5-minute duration, 1 node in the debug queue, stdout+stderr to /home/jinchuli/iri-demo.{out,err}, filesystems home:eagle. `
- `Check the status of the job you just submitted`

### Auth

- `Re-authenticate my ALCF session.`

### Adversarial

- `List the files under `/flare/ChemGraph/jinchu` on Aurora.`
- `Submit a job on Aurora to test whether IRI submission works there yet.`

## Files added

| file | purpose |
|---|---|
| `src/chemgraph/tools/alcf_iri_core.py` | Pure-Python IRI client. Seven action tables (facility, status, account, compute, filesystem, task, auth), the `dispatch()` public API, in-process Globus OAuth for both silent access-token refresh and interactive re-auth. No LangChain deps -- usable from any Python client. |
| `src/chemgraph/tools/alcf_iri_tools.py` | Seven `@tool`-decorated wrappers around `dispatch()`. Mirrors the repo's `*_core.py` + `*_tools.py` split (see cheminformatics, graspa, xanes, pacmof2, ase). |
| `src/chemgraph/graphs/single_agent_iri.py` | LangGraph wiring. Same shape as `graphs/graspa_agent.py`. |
| `src/chemgraph/prompt/alcf_iri_prompt.py` | System prompt for the workflow. |

## Files modified (wiring)

- `src/chemgraph/agent/llm_agent.py` — import, `workflow_map` entry, `single_agent_iri` dispatch branch
- `src/chemgraph/cli/commands.py` — added to `ALL_WORKFLOW_TYPES` + `iri` alias in `WORKFLOW_ALIASES`
- `src/ui/_pages/configuration.py` — added to Streamlit `WORKFLOW_OPTIONS` dropdown

## Design highlights

**Category-plus-discovery pattern.** Six of the seven tools cover ALCF's REST API groups (facility / status / account / compute / filesystem / task). Each accepts `action=<name>` and `params={...}`; three meta-actions (`list_actions`, `describe`, `<any real action>`) let the LLM discover the surface on demand.

**Zero external CLI dependency for auth.** Both silent access-token refresh (Level 1) and interactive browser re-auth (Level 2) use `globus_sdk` in-process with the same public Native App client ID (`8b84fc2d-...`) and scope that ALCF publishes for its `alcf-facility-api-token` helper. The on-disk token cache path is shared, so users who already authenticated via ALCF's CLI helper are unaffected; users who haven't can bootstrap entirely through the in-chat `alcf_auth` tool.

**Two-step OAuth in a chat context.** The `alcf_auth` tool splits the OAuth flow across two tool calls: `start_reauth` returns a Globus URL (agent shows it to the user), `complete_reauth` takes the auth code the user pastes back. No `input()`, no subprocess prompt, works cleanly through Streamlit chat.

**Write actions gated at the tool layer.** `submit_job`, `cancel_job`, `mkdir`, `rm`, `chmod`, etc. raise a clear `RuntimeError` unless `$ALCF_IRI_ALLOW_UNSAFE=1` is set. Tool descriptions explicitly tell the LLM to attempt these normally rather than refuse preemptively — the error surface handles authorization.

**Dynamic resource resolution.** `_resource_id()` calls `GET /status/resources` on first use and caches results per-session. No hardcoded machine-name → UUID fallback table; if ALCF adds a machine or rotates a UUID, the tool picks it up on the next lookup.

## Verified

- Machine status (`get_resource`, `list_resources`, incidents, events) — all 9 currently-registered ALCF resources resolve correctly via live lookup.
- Account (`list_projects`, `list_allocations`) — returns real ChemGraph project + allocation breakdowns.
- Compute (`get_job_status`, `list_jobs`, `submit_job`) — real PBS jobs on Crux (state transitions queued→active→completed).
- Interactive Globus re-auth (`alcf_auth`) end-to-end from a fresh cache: start_reauth → user pastes code → complete_reauth → subsequent tool calls succeed with no restart.
- Silent refresh (`_try_refresh_token`) when access token expires mid-session — invisible recovery, no user interaction.

Not verified in this PR (blocked on ALCF-side identity mapping, tracked in an ALCF support ticket — the code path is correct):
- Filesystem read endpoints (`ls`, `cat`, `head`, ...) — return 422 "Identity failed to map to a local user name" until ALCF ops adds this Globus identity to their identity table. Confirmed working for another user on the same project, so the code is right.
- `stat` — returns 501 "not implemented yet" on ALCF's side.
- Aurora and Sophia `/compute/*` — return 404 "Endpoint resource does not exist for API component compute"; ALCF hasn't wired those clusters into the compute router yet.

## Test plan

- [ ] Fresh `pip install -e .` — verify `chemgraph -w iri -q "is Aurora up"` runs end-to-end.
- [ ] `single_agent_iri` shows in the Streamlit workflow dropdown.
- [ ] Fresh Globus auth via in-chat `alcf_auth`: agent returns a URL, accepts pasted code, subsequent queries succeed.
- [ ] Regression check: other workflows (`single_agent`, `graspa`, `molecular_docking`) unaffected.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/examples/iri/README.md
+++ b/examples/iri/README.md
@@ -1,0 +1,109 @@
+# ALCF IRI Facility API with ChemGraph
+
+Runs the **`single_agent_iri`** workflow that gives an agent access to ALCF's
+IRI Facility API (https://api.alcf.anl.gov). The agent can answer questions
+about ALCF machine state, project allocations, PBS jobs, and (once ALCF
+provisions your identity) remote filesystem contents — all against the real
+API, no ssh required.
+
+You can run it from the CLI:
+```bash
+chemgraph -q "Which ALCF machines are currently up?" -w iri
+```
+or pick **single_agent_iri** as the workflow in the Streamlit UI (`streamlit run src/ui/app.py`).
+
+## What's here
+- `run_chemgraph.py` — runs one query end-to-end and prints the answer.
+- (this README)
+
+## Setup
+
+### 1. Install ChemGraph
+```bash
+pip install -e .
+```
+
+`globus-sdk` is a transitive dependency and gets pulled in automatically.
+
+### 2. Get an ALCF IRI access token
+
+Two paths — pick one.
+
+**A. In-chat (recommended for first use).** Ask the agent to run `alcf_auth`:
+```bash
+chemgraph -w iri -q "Authenticate to ALCF."
+```
+The agent returns a Globus URL. Open it, sign in with your ALCF-linked identity
+(`<your-username>@alcf.anl.gov`), and paste the resulting code back into the
+chat. Tokens cache for 30 days, silently refresh thereafter.
+
+**B. ALCF's CLI helper** (equivalent, terminal-based):
+```bash
+mkdir -p ~/tools
+curl -o ~/tools/alcf_facility_api_globus_token.py \
+    https://raw.githubusercontent.com/argonne-lcf/alcf-facility-api-token/refs/heads/main/alcf_facility_api_globus_token.py
+python ~/tools/alcf_facility_api_globus_token.py authenticate
+export ALCF_API_TOKEN=$(python ~/tools/alcf_facility_api_globus_token.py get_access_token)
+```
+
+Either path writes to the same on-disk token cache
+(`~/.globus/app/8b84fc2d-.../alcf_facility_api_app/tokens.json`),
+so they interoperate.
+
+### 3. LLM backend
+
+Whatever ChemGraph supports (`OPENAI_API_KEY`, `ANTHROPIC_API_KEY`,
+`ARGO_USER` + argo-shim, etc.). See top-level README for options.
+
+## Run
+```bash
+python run_chemgraph.py
+```
+
+## Try your own
+- Public endpoints (machine status, incidents, events, facility metadata)
+  work without any auth token.
+- Authenticated endpoints (projects, allocations, PBS jobs) need step 2 above.
+- Write actions (submit_job, cancel_job, mkdir, rm, chmod, ...) additionally
+  require `export ALCF_IRI_ALLOW_UNSAFE=1`. Use with care.
+
+## Example queries
+
+**Machine status (one hop, no auth)**
+- *Which ALCF machines are currently up?*
+- *Is Aurora in maintenance?*
+
+**Allocations (two hop)**
+- *Find the ChemGraph project and list all its allocations across every machine.*
+- *Which of my allocations is closest to being exhausted?*
+
+**Jobs**
+- *Show the status of Crux job 246115 and describe what happened.*
+- *List my last 5 completed jobs on Polaris.*
+
+**Multi-tool health check (three hop)**
+- *For every compute machine that's up, tell me whether my project has
+  allocation left and how many jobs I have in the queue.*
+
+**Job submission (requires `ALCF_IRI_ALLOW_UNSAFE=1`)**
+- *Submit a smoke-test job on Crux: run `/bin/bash -lc 'echo hello; sleep 10'`
+  under the ChemGraph account, 5-minute duration, 1 node in the debug queue,
+  stdout+stderr to `/home/<user>/iri-demo.{out,err}`, filesystems `home:eagle`.
+  Then keep checking status until it finishes.*
+
+## Design at a glance
+
+Instead of binding one `@tool` per endpoint (~43 tools, ~8k tokens of schemas
+upfront), `single_agent_iri` bundles the endpoints into **7 category tools**
+(facility / status / account / compute / filesystem / task / auth), each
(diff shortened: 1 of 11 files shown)


src/chemgraph/prompt/alcf_iri_prompt.py
```diff
@@ -0,0 +1,35 @@
...
+Each tool takes an `action` and an optional `params` dict. Discovery
+protocol:
+
+  1. If unsure what an action does, call action='describe',
+     params={'target_action': <name>} to get the full schema.
+  2. If you don't know the action name, call action='list_actions'
+     to see all actions the tool supports.
+  3. Then invoke: action=<name>, params={<the params>}.
+
+Machine names ('crux', 'aurora', 'polaris', 'sophia', 'sirius') are
```
tdpham2, line 24: What's sirius?

JinchuLi2002: @tdpham2 It's one of the ALCF resources shown from IRI API:

```
curl -s "https://api.alcf.anl.gov/api/v1/status/resources" | python3 -c "
import json, sys
for r in json.load(sys.stdin):
    if r['name'] == 'Sirius':
        print(json.dumps(r, indent=2))
"
```

returns:
```
{
  "id": "2fef6c82-a982-491e-91a3-d75b59b707af",
  "name": "Sirius",
  "description": "Sirius is an ALCF staging cluster.",
  "last_modified": "2025-07-10T11:28:40.969000Z",
  "group": "computes",
  "current_status": "unknown",
  "resource_type": "compute",
  "self_uri": "https://api.alcf.anl.gov/api/v1/status/resources/2fef6c82-a982-491e-91a3-d75b59b707af",
  "site_uri": "https://api.alcf.anl.gov/api/v1/facility/sites/afb8a6d7-cd5b-4040-8be7-6e917b30af08",
  "capability_uris": []
}
```

examples/iri/README.md
```diff
@@ -0,0 +1,137 @@
...
+(`<your-username>@alcf.anl.gov`), and paste the resulting code back into the
+chat. Tokens cache for 30 days, silently refresh thereafter.
+
+**B. ALCF's CLI helper** (equivalent, terminal-based):
+```bash
+mkdir -p ~/tools
+curl -o ~/tools/alcf_facility_api_globus_token.py \
+    https://raw.githubusercontent.com/argonne-lcf/alcf-facility-api-token/refs/heads/main/alcf_facility_api_globus_token.py
+python ~/tools/alcf_facility_api_globus_token.py authenticate
+export ALCF_API_TOKEN=$(python ~/tools/alcf_facility_api_globus_token.py get_access_token)
```
tdpham2, line 74: Can we add this as an automatic method prior to running any IRI tools? E.g all tools would call alcf_auth prior to their main functionality.

We shouldn't have to authenticate every call, but have a method to check token validity. For example:

```
def alcf_job():
    valid = check_globus_token_valid()
    if not valid:
        alcf_auth() # User validate here if no token is available
    """
    rest of alcf_job code
    """

```

JinchuLi2002: I think it might not be necessary because if the token is not valid, the call will fail with the clear reason being the token issue? 

If we add a helper specifically for checking token validity, we will still need to keep the auth() as a separate tool because the latter requires the user to manually go to the URL and paste back code anyways.

PR #182, opened 2026-08-17
Add MINERVA cluster support for ALCF Inference Endpoints

InkedWings:
## Summary

ALCF base URL resolution assumed every model was served by Sophia, so models on
other ALCF clusters were unreachable.

This adds `ALCF_MINERVA_BASE_URL` and a `supported_alcf_minerva_models` list,
and selects the base URL per model in both `load_alcf_model` and the two
config resolvers. Two Minerva-hosted models are now selectable:
`nemotron-3-ultra` and `inkling-bf16`.

Resolution order is unchanged for every existing model. The new precedence is:

```
--base-url  >  the serving cluster's URL  >  [api.alcf] base_url  >  Sophia
```

The Minerva check deliberately precedes the generic ALCF branch in
`config_utils`. `[api.alcf] base_url` ships pre-set to Sophia in `config.toml`,
the docs, and the UI defaults, and the CLI / eval / UI all pass that resolved
value down as an explicit `base_url` — so without the earlier check it would
shadow the per-cluster fallback and every Minerva request would still go to
Sophia. An explicit `--base-url` still overrides everything.

Note that the two Minerva model IDs appear both in `supported_alcf_models`
(which drives provider detection, so all existing membership checks keep
working untouched) and in `supported_alcf_minerva_models` (which only selects
the URL). `test_minerva_models_are_a_subset_of_supported_alcf_models` guards
against the two lists drifting apart.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Chore / refactor / CI

## How was this tested?

New `tests/test_alcf_base_url.py` (8 cases), covering:

- a Minerva model resolves to the Minerva URL, a Sophia model to the default
- both config resolvers keep the Minerva URL even when `[api.alcf] base_url` /
  `api_alcf_base_url` is set to Sophia
- Sophia models still honour a configured ALCF URL, and fall back to the
  default when none is set
- `load_alcf_model` picks the serving cluster's URL, and an explicit
  `base_url` overrides it
- the Minerva list stays a subset of `supported_alcf_models`

## How was this tested?

New `tests/test_alcf_base_url.py` (8 cases), covering:

- a Minerva model resolves to the Minerva URL, a Sophia model to the default
- both config resolvers keep the Minerva URL even when `[api.alcf] base_url` /
  `api_alcf_base_url` is set to Sophia
- Sophia models still honour a configured ALCF URL, and fall back to the
  default when none is set
- `load_alcf_model` picks the serving cluster's URL, and an explicit
  `base_url` overrides it
- the Minerva list stays a subset of `supported_alcf_models`

Gates run locally:

```bash
ruff check .                       # All checks passed
pytest tests/ -k "not tblite"      # 414 passed, 28 skipped
```

Not covered: no live request was made to the Minerva endpoint from CI or from
my machine, so the routing is verified only down to the resolved URL, not
against a real inference response.

## Checklist

- [x] `ruff check .` passes
- [x] `pytest tests/ -k "not tblite"` passes (plus extras tests if Academy/backen>

---

Written with AI assistance (Claude Code); reviewed before submitting.


tdpham2:
Thanks @InkedWings. This looks nice. For this PR, I think we should also update ALCF model routing in ChemGraph. We should add a prefix to all the ALCF model names (similar to ARGO models) to provide automatic routing. To use ALCF models with ChemGraph, user should add a prefix `alcf:` and we route the llm calls to the right machine (Sophia, Metis and Minerva). I think they already differs the model names on each machine, so this will be useful for us.

For example: 

`meta-llama/Llama-3.3-70B-Instruct` --> `alcf:meta-llama/Llama-3.3-70B-Instruct` (Sophia)
`inkling-bf16` --> `alcf:inkling-bf16` (Minerva)
`Mistral-Large-3-675B-Instruct-2512` -> `alcf:Mistral-Large-3-675B-Instruct-2512` (Metis)

There is a function to normalize Argo models that you can mirror to process ALCF models.

InkedWings:
The `alcf:` prefix and model routing have been implemented in the following commit.

Since `main` now has updated docs, keeping the doc changes in this PR would require a rebase and force push. I think we can resolve the conflicts by accepting the incoming changes and dropping the doc changes from this PR. Then I can open a separate PR to update the content based on the latest version of the docs.

PR #183, opened 2026-08-17
Add OCSR core, model registry, and agreement calibration

reowszer:
## Summary

Optical Chemical Structure Recognition turns a molecule supplied as a figure, scan
or screenshot into a SMILES. Without it the structure is transcribed by hand, which
is slow and error-prone for complex diagrams.

This PR does not install or run any OCSR model: no model environments, no
subprocesses, and no agent-facing tools. Those come in PRs 2 and 3. It contains only
the pure core layer, so the package imports and its tests collect without any
specialist model environments installed.

Agreement scores are empirically calibrated against 722 labelled benchmark images.
For each agreement bucket the table records how often the model majority was
correct, as k, n and a 95% Jeffreys interval. A bucket below n=20 gets a label and
no decimal, because even an apparently strong 18/20 has an interval spanning about
26 percentage points.

The model registry is data-driven. `ocsr_registry.json` defines the models, so
adding one is a JSON edit. Accuracy stays out of it and lives in the calibration
table beside the k and n that produced it.

## Related issues

None.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Chore / refactor / CI

## How was this tested?

On this branch:

- `pytest tests/ -k "not tblite"`: 449 passed, 29 skipped, 2 deselected
  (70 new tests).
- `ruff check .`: passes.

Baseline `main` at c598807 has 19 pre-existing failures, in the durable main-agent
and memory modules. Merging c598807 into this branch reproduces the same 19
failures and no additional ones.

The shipped calibration table was rebuilt from the raw benchmark predictions and
reproduces every bucket's k and n exactly, independently of the code that reads it.

## Part of a 5-PR series

This is the first of five PRs that together add an OCSR capability. They are
intended to land in this order, each building on the last:

1. `feature/ocsr-core` (this PR): core types, model registry, and agreement
   calibration.
2. `feature/ocsr-specialist-envs`: installs the four specialist models.
3. `feature/ocsr-tools`: runs them as subprocesses behind two agent-facing tools.
4. `feature/ocsr-calibrate`: fits a confidence table from user-supplied images.
5. `feature/ocsr-workflow`: registers an `ocsr` workflow.

Split per CONTRIBUTING's guidance on incremental series: the whole feature is
about 6,200 lines and would be difficult to review as a single PR.

## Checklist

- [x] Branched off `main` and targets `main`
- [x] PR is focused on a single logical change
- [x] `ruff check .` passes
- [x] `pytest tests/ -k "not tblite"` passes
- [x] Added tests for the change
- [ ] Updated docs / README: deferred to PRs 3 and 5, where the user-facing
      surfaces land

### DIFF
--- a/src/chemgraph/tools/ocsr_core.py
+++ b/src/chemgraph/tools/ocsr_core.py
@@ -1,0 +1,924 @@
+"""Pure-Python OCSR helpers (no LangChain / MCP decorators).
+
+Optical Chemical Structure Recognition: read a molecule's 2D structure diagram
+from an image and return its SMILES. Used by the LangChain ``@tool`` wrappers in
+:mod:`chemgraph.tools.ocsr_tools`.
+
+Two families of backend. A vision LLM reads the picture directly; local specialist
+models (MolNexTR, MolScribe, DECIMER, OCSRGlyph) are purpose-built image-to-SMILES
+networks driven as subprocesses. Running several specialists and looking at how much
+they agree yields a calibrated confidence, which is the point of :func:`vote` and
+:func:`confidence`.
+
+RDKit is a core dependency and is imported lazily anyway, so this module imports and
+its tests collect on a host with no specialist environment and no network. A
+:func:`mock_ocsr` helper provides deterministic output for hermetic tests.
+"""
+
+from __future__ import annotations
+
+import base64
+import json
+import logging
+import math
+import os
+import re
+import stat
+
+logger = logging.getLogger(__name__)
+
+# Reject anything that is not one of these before base64-ing it to a remote endpoint.
+# Sniffed from the leading bytes, never from the file extension: the extension is
+# attacker-controlled and the whole point is to refuse a non-image renamed to .png.
+_MAGIC = {
+    b"\x89PNG\r\n\x1a\n": "image/png",
+    b"\xff\xd8\xff": "image/jpeg",
+    b"GIF87a": "image/gif",
+    b"GIF89a": "image/gif",
+}
+_MAX_IMAGE_BYTES = 8_000_000
+
+# Confidence label cut-points, applied to the point estimate. Below the calibration
+# table's n floor these are prefixed "low_n_", because a 30-60 pp interval does not
+# deserve a decimal.
+_LABEL_BANDS = ((0.99, "unanimous"), (0.95, "strong"), (0.70, "weak"))
+
+# RDKit parsing is superlinear in string length: 20k characters costs about 9 s of
+# CPU and a megabyte-scale string is effectively unbounded. Model output is untrusted,
+# so cap it before RDKit sees it. Real SMILES are well under 1000 characters; the
+# longest in the OCSR benchmark is 224.
+_MAX_SMILES_CHARS = 4000
+
+_SMILES_LABEL_RE = re.compile(r"^\s*(?:smiles|answer|result)\s*[:=]\s*", re.IGNORECASE)
+
+
+# ---------------------------------------------------------------------------
+# Image loading
+# ---------------------------------------------------------------------------
+
+
+def _sniff_mime(head: bytes) -> str | None:
+    """Return the MIME type implied by a file's leading bytes, or None.
+
+    WEBP needs a split check: "RIFF" then "WEBP" four bytes later.
+    """
+    for magic, mime in _MAGIC.items():
+        if head.startswith(magic):
+            return mime
+    if head[:4] == b"RIFF" and head[8:12] == b"WEBP":
+        return "image/webp"
+    return None
+
+
+def load_image_bytes(image_path: str, max_bytes: int = _MAX_IMAGE_BYTES) -> tuple[bytes, str]:
+    """Read an image file safely and return ``(raw_bytes, mime_type)``.
+
+    Validates before reading rather than after. Reading first and checking later is
+    not equivalent: ``/dev/zero`` grows the buffer without bound, a FIFO blocks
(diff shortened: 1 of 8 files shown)


src/chemgraph/tools/ocsr_core.py
```diff
@@ -0,0 +1,924 @@
...
+    classifier.
+    """
+    if " " in s.strip():
+        return True
+    # A SMILES is mostly non-alphanumeric structure or organic-subset letters. A word
+    # like "images" is all lowercase letters and no structural characters at all.
+    return s.isalpha() and s.islower() and not set(s) <= set("cnopsbrifhe")
+
+
+def extract_smiles(raw_text: str | None) -> str | None:
```
tdpham2, line 275: I don't think this can catch and filter all different LLM answers. Example failure could be SMILES include in markdown (** **), SMILES in paranthesis ( ). 

What if we do structured output here (route to a separate LLM call), or simply put the JSON structure output in the LLM's prompt?

src/chemgraph/tools/ocsr_registry.json
```diff
@@ -0,0 +1,132 @@
...
+  "schema": 1,
+  "_comment": "Registry of OCSR models. To wire in a new specialist: add an entry under 'specialists' and drop its worker script into chemgraph/tools/ocsr_workers/. No Python change is needed -- the backend, the installer, the listing, and ensemble membership are all derived from this file. Measured accuracy lives in the calibration table, not here, next to the k and n that produced it; run chemgraph.tools.ocsr_calibrate to fit one.",
+  "specialists": {
+    "decimer": {
+      "latency_s": 0.73,
+      "note": "Most accurate single model. The default for backend='auto'.",
+      "worker": "decimer_infer.py",
+      "install": {
+        "script": "build_decimer.sh",
+        "weights": "~/.data/DECIMER-V2",
```
tdpham2, line 11: Are the envs/weights autoamatically created when installing the models, or invoking it the first time?

src/chemgraph/tools/ocsr_registry.json
```diff
@@ -0,0 +1,132 @@
+{
+  "schema": 1,
+  "_comment": "Registry of OCSR models. To wire in a new specialist: add an entry under 'specialists' and drop its worker script into chemgraph/tools/ocsr_workers/. No Python change is needed -- the backend, the installer, the listing, and ensemble membership are all derived from this file. Measured accuracy lives in the calibration table, not here, next to the k and n that produced it; run chemgraph.tools.ocsr_calibrate to fit one.",
+  "specialists": {
+    "decimer": {
+      "latency_s": 0.73,
+      "note": "Most accurate single model. The default for backend='auto'.",
+      "worker": "decimer_infer.py",
+      "install": {
+        "script": "build_decimer.sh",
```
tdpham2, line 10: Is/Should the build script be bundled with ChemGraph?

src/chemgraph/tools/ocsr_core.py
reowszer: Thanks for the review. Both examples reproduce, and the markdown case exposed a more serious
failure mode: `*` is RDKit's wildcard atom, so `**CCO**` parses as a valid five-atom
molecule and was returned as the prediction. `(CCO)` failed to parse.

Fixed in `29e8a79`. Emphasis, quotes and balanced brackets are peeled before parsing,
and brackets come off only when the wrapped form fails to parse, so `[NH4+]` and
`C(=O)O` survive. A `"smiles"` key is read first when the reply is JSON.

Exact-match extraction improved from 26.8% to 37.9%, with 15 new tests covering
these cases.

The prompt suggestion is also addressed in #185. `prompt_mode="structured"` asks
for `{"smiles": "..."}`. Both modes were kept because the shipped calibration table
was measured with the legacy prompt; a structured run should refit with
`ocsr_calibrate` (#186).

The existing inference path was kept. A second LLM call would add one network round
trip per prediction and four per ensemble call. The parsing path is required for the
four specialist models, whose interface is `predict_SMILES(image)` with no prompt.

src/chemgraph/tools/ocsr_registry.json
reowszer: Neither happens implicitly. Both are created by the explicit build script in #184.

`python -m chemgraph.tools.ocsr_setup decimer` runs `build_decimer.sh`, which creates
the conda env at `~/ocsr-decimer`. A warm-up inference then downloads the Zenodo
weights into `~/.data/DECIMER-V2`, so the first real call does not pay for the
download.

If a model is not installed, the tool returns a structured error naming the model and
the exact install command. Sizes are in the registry so the cost is visible before
starting: DECIMER is 2.6 GB of environment plus 0.66 GB of weights.

src/chemgraph/tools/ocsr_registry.json
reowszer: Yes. As of #184 the four `build_*.sh` scripts live in
`src/chemgraph/tools/ocsr_workers/` and ship as package data through
`"chemgraph.tools.ocsr_workers" = ["*.py", "*.sh", "*.txt"]` in `pyproject.toml`, so
they are present in an installed wheel and `ocsr_setup` finds them with
`importlib.resources`.

They are bundled because each one encodes a specific dependency resolution that is
easy to get subtly wrong: MolScribe needs torch 1.13 built against numpy 1.x, which
conflicts with ChemGraph's own numpy 2.2.6 pin. Keeping them next to the registry
also lets `ocsr_setup --check` tell "this model is not installed" apart from "this
registry names a file the package does not contain".

Happy to move them to a top-level `scripts/` directory if you would prefer them out
of the package.

PR #192, opened 2026-08-18
fix: improve ASE result metadata

### LINKED ISSUE #190: [REFACTOR] Update run_ase tool to emit model name
In run_ase tool for mace calculations, when model_name=none, run_ase_core function call the default model for each calculator_type (mace_mp, mace_off). However, when writing the output to JSON file, the model_name is still written as null, causing agent to sometimes run the optimization in a loop.

tdpham2:
## Summary

- Record the maintained default MACE model name in successful ASE output when callers omit `model`, while continuing to pass `None` to MACE itself.
- Add `potential_energy` as the canonical ASE energy field and retain `single_point_energy` as a deprecated compatibility alias.
- Report the driver, energy unit, results path, convergence state, and optimization step count consistently.
- Use driver-aware completion messages and report labels, with legacy-energy fallback in the Parsl ensemble aggregator.

Fixes #190.

## Compatibility

- Existing consumers can continue reading `single_point_energy`.
- Explicit MACE model selections are unchanged.
- The output-only default model mapping records `medium-mpa-0` for `mace_mp` and `medium` for `mace_off`.

## Testing

- `ruff check .`
- Focused `.cg_fix_ase` CI regression tests: 5 passed
- Full `.cg_fix_ase` suite with `pytest tests/ -k "not tblite"`: 449 passed, 29 skipped, 2 deselected

## Checklist

- [x] Added regression coverage for default MACE model persistence
- [x] Added regression coverage for energy metadata and optimization completion
- [x] Preserved backward compatibility for existing result consumers
- [x] Kept behavior changes in separate topic-focused commits

### DIFF
--- a/src/chemgraph/mcp/mace_mcp_parsl.py
+++ b/src/chemgraph/mcp/mace_mcp_parsl.py
@@ -16,7 +16,7 @@ from chemgraph.schemas.mace_parsl_schema import (
     mace_input_schema,
     mace_input_schema_ensemble,
 )
-from chemgraph.tools.parsl_tools import run_mace_core
+from chemgraph.tools.parsl_tools import _result_potential_energy, run_mace_core
 from parsl import python_app
 
 warnings.warn(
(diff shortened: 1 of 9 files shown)


tdpham2:
Thanks @InkedWings for running evaluation for this.

PR #193, opened 2026-08-18
Add OpenRouter as a first-class provider

### LINKED ISSUE #168: [FEATURE] Expand ChemGraph's LLM providers
ChemGraph currently supports major LLM providers (OpenAI, Anthropic, etc.), through API-key-based authentication.

Codex subscription support was recently added in #167. We should extend this approach to support additional subscription-based providers where technically feasible, particularly:

- Hugging Face subscriptions for accessing open-weight models
- GitHub Copilot

We should also add support for additional API-based providers, especially OpenRouter, to broaden the range of models available for ChemGraph evaluation.

InkedWings:
## Summary

Adds [OpenRouter](https://openrouter.ai) as a provider, selected with an
`openrouter:` prefix in the same style as the existing `groq:` and `codex:`
routes. OpenRouter is OpenAI-compatible, so the loader returns a `ChatOpenAI`
bound to `https://openrouter.ai/api/v1` and keyed on `OPENROUTER_API_KEY`. The
prefix is stripped before the request goes out, so the slug reaches OpenRouter
verbatim.

`[api.openrouter]` ships in `config.toml`, so OpenRouter coexists with Argo and
ALCF in one config file, and `OPENROUTER_API_KEY` is forwarded through
docker-compose and the k8s manifests alongside the other provider keys.

Dispatch is by prefix, not list membership: `supported_openrouter_models` is a
discovery list for `--list-models` and for the per-model quirk sets, not a gate.
Any slug from openrouter.ai/models works without a release, which matters
because that catalog changes far faster than ChemGraph does. This matches how
`groq:` and `codex:` already behave.

Fix #168 

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Chore / refactor / CI

## How was this tested?

`tests/test_openrouter_provider.py` (new, 187 lines) covers prefix dispatch,
base-URL resolution through both config resolvers, the API-key check, the
non-TTY failure path, and the constructed client parameters. No network access.

Gates:

```bash
ruff check .                     # All checks passed
pytest tests/ -k "not tblite"    # 453 passed, 28 skipped, 2 deselected
```

Also verified end to end against the live API — `openrouter:deepseek/deepseek-v4-flash`
answered a real query through the CLI, confirming the prefix is stripped
correctly and the request reaches openrouter.ai.

The k8s and docker-compose changes were syntax-checked but not deployed.

## Checklist

- [x] `ruff check .` passes

---

Written with AI assistance (Claude Code); reviewed before submitting.

### DIFF
--- a/src/chemgraph/models/openrouter.py
+++ b/src/chemgraph/models/openrouter.py
@@ -1,0 +1,99 @@
+"""Load OpenRouter models using LangChain's OpenAI-compatible client."""
+
+import os
+import sys
+from getpass import getpass
+
+from langchain_openai import ChatOpenAI
+
+from chemgraph.models.supported_models import (
+    MODELS_WITHOUT_TEMPERATURE,
+    OPENROUTER_DEFAULT_BASE_URL,
+)
+from chemgraph.utils.logging_config import setup_logger
+
+logger = setup_logger(__name__)
+
+# Higher than the OpenAI/Groq defaults on purpose: most models served through
+# OpenRouter emit reasoning tokens by default, and those count against the
+# *completion* budget. Too small a cap lets chain-of-thought exhaust it before
+# the tool call is emitted, producing an empty turn the graph reads as a dead end.
+OPENROUTER_DEFAULT_MAX_TOKENS = 8000
+
+
+def load_openrouter_model(
+    model_name: str,
+    temperature: float = 0.0,
+    api_key: str = None,
+    base_url: str = None,
+    max_tokens: int = OPENROUTER_DEFAULT_MAX_TOKENS,
+) -> ChatOpenAI:
+    """Load an OpenRouter chat model into LangChain.
+
+    OpenRouter exposes an OpenAI-compatible API, so this returns a
+    ``ChatOpenAI`` pointed at the OpenRouter endpoint. Any slug from
+    https://openrouter.ai/models works; ``supported_openrouter_models`` is a
+    discovery list, not a gate.
+
+    Parameters
+    ----------
+    model_name : str
+        Model identifier with the ``openrouter:`` prefix. The prefix is
+        stripped and the remainder sent verbatim (e.g. ``moonshotai/kimi-k3``).
+    temperature : float
+        Sampling temperature. Omitted for ``MODELS_WITHOUT_TEMPERATURE`` entries.
+    api_key : str, optional
+        Falls back to ``OPENROUTER_API_KEY``. Never to ``OPENAI_API_KEY``.
+    base_url : str, optional
+        Endpoint override. Defaults to ``OPENROUTER_DEFAULT_BASE_URL``.
+    max_tokens : int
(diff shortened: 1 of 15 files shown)


src/chemgraph/models/openrouter.py
```diff
@@ -0,0 +1,99 @@
...
+                "environment variable:\n"
+                "  export OPENROUTER_API_KEY='your_key_here'\n"
+                "  Get a key at: https://openrouter.ai/keys"
+            )
+
+    logger.info("Loading OpenRouter model: %s", model_name)
+    llm_kwargs = dict(
+        model=model_name,
+        api_key=api_key,
+        base_url=base_url or OPENROUTER_DEFAULT_BASE_URL,
```
tdpham2, line 88: Should we fix this to `OPENROUTER_DEFAULT_BASE_URL`?

I'm thinking if one uses openrouter:model, and provides base_url, our implementation isn't needed. Since your stripped `openrouter:`, this will point to a completely different provider. Want to know your thoughts on this.

src/chemgraph/models/supported_models.py
```diff
@@ -121,6 +121,21 @@
...
+# Default OpenRouter API base URL (OpenAI-compatible).
+OPENROUTER_DEFAULT_BASE_URL = "https://openrouter.ai/api/v1"
+
+# OpenRouter models -- use the "openrouter:" prefix (e.g.
+# "openrouter:moonshotai/kimi-k3"). The prefix is stripped before the request
+# and the remainder is sent as the OpenRouter slug verbatim.
+# This list is curated for *discovery* (--list-models, UI dropdown, per-model
+# quirk sets such as MODELS_WITHOUT_TEMPERATURE); it is NOT a gate. Dispatch is
+# by prefix, so any slug from https://openrouter.ai/models also works.
+supported_openrouter_models = [
```
tdpham2, line 133: Can you add more modern models to this list? Maybe all the more recent kimi, deepseek, qwen, GLM.

Also make sure to update MODELS_WITHOUT_TEMPERATURE list.

src/chemgraph/models/openrouter.py
tdpham2: Actually I also see your changes in config.toml so this is probably fine.

src/chemgraph/models/supported_models.py
InkedWings: I added recent models from DeepSeek, Kimi, Qwen, GLM, and MiniMax, and ran one end-to-end test for each of them. All of these models support setting temperature, so they don’t need to be added to MODELS_WITHOUT_TEMPERATURE.

I also updated the previous comments since they were a bit misleading.

keceli:
I was able to use OpenRouter by `-m openai/gpt-4o --base-url https://openrouter.ai/api/v1` with api key provided in `OPENAI_API_KEY` last week. Is there any additional capability the PR provides, maybe support for models not compatible with OpenAI?

tdpham2:
LGTM! Thanks @InkedWings

PR #195, opened 2026-08-19
Add tool and worker agent registries

tdpham2:
## Summary

- Add a lazy `ToolRegistry` for ChemGraph’s 20 individually exported in-process LangChain tools, with metadata, tag filtering, custom registration, and runtime availability checks.
- Add an `AgentRegistry` for the ten worker workflows, including aliases, standalone construction, and `CompiledSubAgent` adaptation with inherited checkpointing.
- Add registry documentation and a credential-free example that demonstrates discovery, availability checks, tool resolution, and optional worker construction.
- Keep MCP tools, generated IRI API tool collections, skills, benchmarks, selector middleware, and `main_agent` registration out of scope; `main_agent` remains the orchestration graph.

## Related issues

None.

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Docs
- [ ] Chore / refactor / CI

## How was this tested?

- `ruff check .`
- `pytest tests/ -k "not tblite"`: 471 passed, 29 skipped, 2 deselected
- `mkdocs build --strict`
- `PYTHONPATH=src python examples/registries/run_registries.py`

## Checklist

- [x] `ruff check .` passes

### DIFF
--- a/src/chemgraph/registry/agents.py
+++ b/src/chemgraph/registry/agents.py
@@ -173,24 +173,29 @@ class AgentRegistry:
             self.register(spec)
 
     def register(self, spec: AgentSpec, *, replace: bool = False) -> None:
-        """Register one worker specification."""
+        """Register or replace one worker specification."""
         if not isinstance(spec, AgentSpec):
             raise TypeError("AgentRegistry entries must be AgentSpec objects.")
         if not spec.name:
             raise ValueError("Registry names must not be empty.")
 
+        previous = self._specs.get(spec.name)
         claimed = (spec.name, *spec.aliases)
-        duplicates = [
-            name
-            for name in claimed
-            if name in self._specs or name in self._aliases
-        ]
-        if duplicates and not (replace and duplicates == [spec.name]):
-            raise DuplicateRegistryEntryError(
-                f"Agent name or alias already registered: {duplicates!r}."
+        collisions = []
+        for name in claimed:
+            owner = name if name in self._specs else self._aliases.get(name)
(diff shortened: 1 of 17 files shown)


tdpham2:
Next step is to update the existing tools and subgraphs before using it in the main_agent graph.

src/chemgraph/registry/agents.py
```diff
@@ -0,0 +1,363 @@
...
+        if not spec.name:
+            raise ValueError("Registry names must not be empty.")
+
+        claimed = (spec.name, *spec.aliases)
+        duplicates = [
+            name
+            for name in claimed
+            if name in self._specs or name in self._aliases
+        ]
+        if duplicates and not (replace and duplicates == [spec.name]):
```
keceli: Codex found this: `AgentRegistry.register(..., replace=True)` currently rejects replacing an entry if the replacement keeps one of the existing aliases, because duplicate detection runs before the old aliases are removed. That makes the documented “intentionally replaces” path a bit narrower than it sounds.

I think either fix is fine:
- update `register()` so `replace=True` allows names/aliases already owned by the same canonical agent, while still rejecting aliases owned by other entries; or
- narrow the docs to say replacement only supports the canonical name and aliases must be changed/removed first.

Not blocking core built-in registry behavior, but worth aligning the API/docs before this becomes public surface.

tdpham2: I fixed this in the latest commit

keceli (approved):

PR #198, opened 2026-08-19
Add image-to-SMILES tools with four local models and an LLM fallback

reowszer:
## Summary

ChemGraph has no way to read a chemical structure diagram. A user with a figure from a paper transcribes the SMILES by hand before any other tool applies.

This adds `image_to_smiles(image_path, model=...)`, where `model` is one of `decimer`, `molnextr`, `molscribe`, `ocsrglyph`, or `llm`, plus an `ocsr` workflow that binds it and a runnable example. It replaces #183 to #187.

**Default: DECIMER.** Highest exact match of the four (0.899 over a 722-image benchmark) and the only one whose published package installs into ChemGraph's environment without overriding its dependencies, satisfying both criteria raised in review.

**Model selection.** Unset takes the most accurate installed specialist, or `llm` when none of the four is installed. A named model is never silently swapped: it runs, or it reports why it cannot.

**LLM fallback uses the agent's own model.** `make_ocsr_tools(llm)` closes over the chat model the agent is already running, so the tool resolves no credentials of its own.

**Checkpoints are checked before use.** The error names the missing path instead of failing inside the model. DECIMER fetches and caches its own through pystow. `CHEMGRAPH_OCSR_WEIGHTS_DIR` sets where the tool looks for the other three, for clusters where the weights live on shared scratch.

### Installation

All four models run in ChemGraph's own environment, in one process, with no change to its pinned versions.

```bash
pip install 'chemgraph[ocsr]'
```

That installs DECIMER, the default, which is enough to use the tool.

The other three install from source, with the steps in `examples/ocsr/README.md`. MolNexTR and OCSRGlyph are unpublished on PyPI. MolScribe is published, but that release pins `torch<2.0` and `numpy<2.0`, which pip would resolve by downgrading both below the versions ChemGraph pins. The extra can cover all three once each has an installable package with current pins.

No `build_*.sh` scripts: separate conda environments are no longer needed.

Two of the four declare requirements that suggest they cannot share this environment. Each was checked against the installed source:

| model | published requirement | what holds |
|-------|----------------------|------------|
| MolNexTR | `timm==0.4.12` | timm 1.x through the shim |
| MolScribe | `torch<2.0`, `numpy<2.0`, `timm==0.4.12` | the torch and numpy caps are in the PyPI release only; git main asks `torch>=1.11.0`. timm 1.x through the shim |
| OCSRGlyph | none published | `timm>=1.0`, no conflict |
| DECIMER | `requires-python >=3.9` | satisfied |

### The timm shim

MolNexTR and MolScribe each vendor a Swin Transformer written against timm 0.4.12 internals, and the two copies break on the same imports. `chemgraph/tools/timm_compat.py` restores those paths, so both run on current timm with their own source untouched. It is a no-op on timm 0.4.12, and the backend applies it before loading either model.

With that in place every model loads in-process and no per-model environment is required, so this PR adds no subprocess worker layer.

### Out of scope

No confidence number is reported. Per-image calibrated confidence comes from the committee workflow, which is the follow-up PR along with `ocsr_calibrate` and the vote path.

## Related issues

Replaces #183, #184, #185, #186, #187.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Chore / refactor / CI

## How was this tested?

All four models run end to end through the tool on `examples/ocsr/images/aspirin.png`, in a single process, on Python 3.12 with `numpy==2.2.6`, `torch==2.13.0` and `timm==1.0.28`:

| model | result | formula |
|-------|--------|---------|
| decimer | `CC(=O)OC1=CC=CC=C1C(=O)O` | C9H8O4 |
| ocsrglyph | `CC(=O)Oc1ccccc1C(=O)O` | C9H8O4 |
| molnextr | `CC(=O)Oc1ccccc1C(=O)O` | C9H8O4 |
| molscribe | `CC(=O)Oc1ccccc1C(=O)O` | C9H8O4 |

DECIMER emits the Kekule form and the others the aromatic form; RDKit canonicalizes all four to the same molecule.

The shim touches MolNexTR and MolScribe only, so those two were checked for prediction equivalence. DECIMER is TensorFlow-based and OCSRGlyph already wants timm 1.x, so neither passes through it. With the shim isolated as the only variable, both affected models ran over the same 60 PubChem eval images on timm 1.0.28 with the shim and on their pinned timm 0.4.12 without it, with the timm version as the only change made between a model's two runs; a `pip freeze` diff across that switch confirms it moves no other package.

| model | identical predictions | errors |
|-------|----------------------|--------|
| MolNexTR | 60/60 | 0 |
| MolScribe | 60/60 | 0 |

`examples/ocsr/run_ocsr.py` checks its predictions against the four structures the images were drawn from.

New tests: `tests/test_ocsr_tools.py` (20), `tests/test_timm_compat.py` (7), and `tests/test_ocsr_core.py` reworked for this contract. 71 passed, 1 skipped in the three files.

`ruff check .` passes. `pytest tests/ -k "not tblite"` does not complete locally: twelve modules fail to collect because `deepagents` is not installed. The same twelve fail on `main` at 17bb51e, so this is unrelated to the change. Excluding them, 321 passed and 28 skipped.

## Checklist

- [x] `ruff check .` passes
- [ ] `pytest tests/ -k "not tblite"` passes (see above: blocked locally by a missing `deepagents`, unrelated and reproduced on `main`)


pyproject.toml
```diff
@@ -52,6 +52,16 @@ dependencies = [
     ]
 
 [project.optional-dependencies]
+# Optical Chemical Structure Recognition: read a structure diagram, get a SMILES.
+#
+# DECIMER alone covers the default path and is the only one of the four specialists
+# published on PyPI. The other three are installed from source; see
+# examples/ocsr/README.md, which also covers the checkpoints. All four run in this
+# environment, so none needs one of its own.
+ocsr = [
```
tdpham2, line 62: I see on MolNexTR, MolScribe and glyph, you can do pip for installation:

`pip install git+https://github.com/CYF2000127/MolNexTR`
`pip install MolScribe`
`pip install -e '.[all]'`. <--- Try to point .[all] to their GitHub repo

Can you test if their methods work and add these (also make sure to build a new env to check)? While I'd prefer if they are on PyPI, we don't have to

reowszer: Tested all three installation paths in a fresh Python 3.12 venv. `pip install MolScribe` fails on its `torch<2.0` pin, though `git+.../MolScribe` works. `git+.../MolNexTR` fails on `pyonmttok==1.37.1`, which has no 3.12 wheel; you were right that it has packaging, and I have corrected my note in the PR saying otherwise. `.[all]` on OCSRGlyph works straight from upstream.

Installing them together is the real blocker. MolNexTR and MolScribe pin `timm==0.4.12`, OCSRGlyph needs `timm>=1.0`, and no version satisfies both. Installing them one at a time lets the later install replace the timm version the earlier one needs, and pip reports success while OCSRGlyph is broken at inference:

```
RuntimeError: features_only not implemented for Vision Transformer models
```

So I went with the forks you suggested. Each one changes two lines of dependency metadata and nothing else, and with the extra pointing at them `pip install 'chemgraph[ocsr]'` installs and runs all four, verified end to end in a clean venv. The URLs are my forks for now. That is a separate commit.

The other commit is unrelated to installation. It fixes a segfault when DECIMER runs before any torch specialist, caused by the TensorFlow and torchvision import order, and is independent of the fork decision.

tdpham2: Looks good! I'm running test on my end before merging. 

One additional thing for this. Could you also pin the commit for each repo (in case each repo changes later, we make sure everything will keep working on our end)

src/chemgraph/tools/ocsr_backends.py
```diff
@@ -0,0 +1,373 @@
...
+
+
+def _resolve_weights(name: str) -> tuple[str | None, str]:
+    """Return (path, error). A missing checkpoint is an error naming the fix."""
+    path = checkpoint_path(name)
+    if path is None:
+        return None, ""
+    if os.path.exists(path):
+        return path, ""
+    return None, (f"{name}'s checkpoint is missing at {path}. Fetch it with: "
```
tdpham2: Is this still valid? The chemgraph.tools.ocsr_download isn't committed in this PR.

src/chemgraph/tools/ocsr_core.py
```diff
@@ -0,0 +1,421 @@
...
+    """
+    for magic, mime in _MAGIC.items():
+        if head.startswith(magic):
+            return mime
+    if head[:4] == b"RIFF" and head[8:12] == b"WEBP":
+        return "image/webp"
+    return None
+
+
+def load_image_bytes(image_path: str, max_bytes: int = _MAX_IMAGE_BYTES) -> tuple[bytes, str]:
```
tdpham2, line 91: Codex found this. Can you check?

load_image_bytes (https://github.com/argonne-lcf/ChemGraph/blob/7c2fbf58083de3fec45a32812b0abeb95b6d5a41/src/chemgraph/tools/ocsr_core.py#L100-L115) resolves relative paths against the current directory instead of using _resolve_existing_path. Consequently, a bare image filename under CHEMGRAPH_LOG_DIR fails even though sibling readers support it. Additionally, image_to_smiles_core (https://github.com/argonne-lcf/ChemGraph/blob/7c2fbf58083de3fec45a32812b0abeb95b6d5a41/src/chemgraph/tools/ocsr_tools.py#L86-L97) passes the original path to specialists, discarding tilde/absolute normalization performed during validation.

examples/ocsr/README.md
tdpham2, line 1: Need to update README as we install all 4 packages now.

src/chemgraph/tools/ocsr_backends.py
```diff
@@ -0,0 +1,373 @@
...
+from typing import Any, Callable
+
+# Import torchvision before anything pulls in TensorFlow. Loading it into a process
+# that already holds TensorFlow segfaults inside torchvision.ops, so a session that
+# reads one image with DECIMER and the next with any torch model dies on the second
+# call. The reverse order is safe, and importing torch alone is safe; torchvision is
+# the one that has to go first. Keep this at module scope: the specialists are
+# imported lazily, which is exactly what makes the order accidental otherwise.
+try:
+    import torchvision  # noqa: F401
```
tdpham2: Can you change this to lazy import? The issue is llm_agent import ocsr_agent, which uses ocsr_core and ocsr_backends. So for an installation that don't need torchvision (e.g current workflow), I think this may break the system.

src/chemgraph/tools/ocsr_tools.py
```diff
@@ -0,0 +1,268 @@
...
+        )
+
+    validation = core.validate_smiles_core(narrow["smiles"])
+    warning = ""
+    if validation.get("n_fragments", 0) > 1:
+        warning = (f"the image contains {validation['n_fragments']} disconnected "
+                   f"fragments (a salt, a mixture, or a reaction scheme). Ask which "
+                   f"one is meant before using this SMILES.")
+
+    return core.build_result(
```
tdpham2, line 128: Codex found this. Can you check?

The returned SMILES does not match its documented contract. The tool says smiles is canonicalized, but the success path returns the model’s original string
     (https://github.com/argonne-lcf/ChemGraph/blob/7c2fbf58083de3fec45a32812b0abeb95b6d5a41/src/chemgraph/tools/ocsr_tools.py#L113-L123).

DECIMER’s Kekulé output therefore remains different from another model’s aromatic output for the same molecule. Either return validation["canonical_smiles"]—with an explicit stereochemistry policy or document the field as raw and expose canonical SMILES separately.

src/chemgraph/prompt/ocsr_prompt.py
```diff
@@ -0,0 +1,106 @@
...
+# The structured alternative. Not vendored, and free to change: it is deliberately
+# outside the block above so the legacy benchmark prompt stays byte-identical.
+#
+# A model asked for one bare SMILES still returns markdown, prose and JSON, and
+# extract_smiles has to guess which. Asking for JSON makes the answer a named field.
+# The published accuracies were measured under the benchmark prompt above, so they
+# do not describe a run that uses this one.
+# ---------------------------------------------------------------------------
+
+OCSR_STRUCTURED_SYSTEM_PROMPT = (
```
tdpham2, line 48: This system prompt is not used in other parts of the commit

src/chemgraph/tools/ocsr_core.py
```diff
@@ -0,0 +1,421 @@
...
+    return data, mime
+
+
+def load_image_b64(image_path: str, max_bytes: int = _MAX_IMAGE_BYTES) -> tuple[str, str]:
+    """Read an image and return ``(base64_ascii, mime_type)`` for a data URL."""
+    data, mime = load_image_bytes(image_path, max_bytes=max_bytes)
+    return base64.b64encode(data).decode("ascii"), mime
+
+
+def extract_image_path(text: str) -> str | None:
```
tdpham2: I think this is only used in the test and not other parts of the main code. So probably clean this and the associated test.

src/chemgraph/tools/ocsr_core.py
```diff
@@ -0,0 +1,421 @@
...
+    mime = _sniff_mime(data[:16])
+    if mime is None:
+        raise ValueError(
+            f"not a recognized image (PNG/JPEG/GIF/WEBP) by content: {path}. "
+            f"The extension is not trusted; only the leading bytes are."
+        )
+    return data, mime
+
+
+def load_image_b64(image_path: str, max_bytes: int = _MAX_IMAGE_BYTES) -> tuple[str, str]:
```
tdpham2: I think this is only used in the test and not other parts of the main code. So probably clean this and the associated test. Same as below

pyproject.toml
reowszer: Done. Each of the three now pins the commit it was tested against:

```
"glyph @ git+https://github.com/EdisonScientific/glyph@0bf782f863d26b041ace157668928ef07c38b972",
"MolNexTR @ git+https://github.com/reowszer/MolNexTR@f450b9661557b1f91ae36f59dd1fadfbcb3a0967",
"MolScribe @ git+https://github.com/reowszer/MolScribe@b03b30fbac9a78434116e626ebef6c7b7bdcdb6e",
```

Each fork is two changed lines of `setup.py` against upstream and nothing else. The same commit tightens the other two entries: `decimer` was unconstrained, and `timm` is now capped below 1.1 because `timm_compat` uses its private modules.

src/chemgraph/tools/ocsr_backends.py
reowszer: Good catch, it is not. That module was part of an earlier design and the message survived the rewrite, so it pointed at a command that does not exist. The error now names examples/ocsr/README.md and CHEMGRAPH_OCSR_WEIGHTS_DIR, and the README gained the actual download commands for the three checkpoints.

src/chemgraph/tools/ocsr_core.py
reowszer: Both correct, both fixed. The path is resolved once before the format check, and the resolved path is what the specialist opens, so it reads the file that was validated.

One deviation from the suggestion: the three lines are inline instead of imported from ase_core. That module reaches the fairchem calculator schema, which imports torch at module scope, so importing it would undo the lazy import in your last comment. Tests cover the log-dir lookup and cwd precedence.

examples/ocsr/README.md
reowszer: Updated. The install section is now the one extra, with the reason the two forks exist. The per-model install column is gone from the table. I also added the Hugging Face commands for the three checkpoints, since the missing-checkpoint error now points there.

src/chemgraph/tools/ocsr_backends.py
reowszer: You are right. llm_agent reaches ocsr_agent on every workflow, so an install without torch would have failed on workflows that never touch OCSR.

The preload now runs just before a specialist loads, which is the only point where the import order matters. Verified both ways: with torch, torchvision and TensorFlow blocked the OCSR modules import and `image_to_smiles_core` returns its error result, and in the full environment DECIMER followed by MolScribe still runs clean. A test covers the blocked-import case, and it is what caught an `ase_core` import chain that the path-resolution fix would otherwise have reintroduced.

src/chemgraph/tools/ocsr_tools.py
reowszer: Confirmed, and I took the first option with one change. `validation["canonical_smiles"]` is stereo-stripped to match how the benchmark scores, so returning it would drop stereochemistry the model read correctly. Penicillin G loses all four stereocentres:

```
validation["canonical_smiles"]  CC1(C)SC2C(NC(=O)Cc3ccccc3)C(=O)N2C1C(=O)O
returned by the tool            CC1(C)S[C@@H]2[C@H](NC(=O)Cc3ccccc3)C(=O)N2[C@H]1C(=O)O
```

The return site therefore canonicalizes separately, with stereochemistry kept. An unparseable string is returned raw with `valid=False`, so this cannot blank an answer, and all four models now return the same string for the aspirin image where DECIMER previously returned the Kekule form. A test fails if the field is ever switched to `validation["canonical_smiles"]`.

src/chemgraph/prompt/ocsr_prompt.py
reowszer: Now it is. You asked for structured output on this path in #183:

> What if we do structured output here (route to a separate LLM call), or simply put
> the JSON structure output in the LLM's prompt?

The prompt was written for that, and the rewrite that collapsed the LLM backend dropped the argument selecting it. `structured=True` sends it again: the reply comes back as `{"smiles": ...}`, read from a named field, and an image that is not a molecule can answer with null. It is on the tool schema, so an agent can set it after a plain call comes back unreadable.

The default still sends the vendored prompt, since that is the one the published accuracies were measured under, and free-text parsing is untouched. `OCSR_SYSTEM_PROMPT` and `OCSR_USER_PROMPT` stay: the `model="llm"` path sends both.

src/chemgraph/tools/ocsr_core.py
reowszer: Both removed. `extract_image_path` had a test of its own, which went with it; `load_image_b64` was covered by two assertions inside the `load_image_bytes` test, so only those came out and that test stays.

src/chemgraph/tools/ocsr_core.py
reowszer: Both removed. `extract_image_path` had a test of its own, which went with it; `load_image_b64` was covered by two assertions inside the `load_image_bytes` test, so only those came out and that test stays.

PR #199, opened 2026-08-20
feat(iri): MCP server + skill card for ALCF IRI tools

JinchuLi2002:
## Summary

Follows up merged PR #173:

- **MCP server** for the ALCF IRI Facility API: `chemgraph.mcp.alcf_iri_mcp` — 43 flat tools, one per endpoint, launchable via `python -m chemgraph.mcp.alcf_iri_mcp` (stdio) or `--transport streamable_http --port 9010`. Wraps the same `alcf_iri_core` implementation as `single_agent_iri`, so auth / refresh / unsafe-write gating / error surface are identical. Any MCP-speaking client (Claude Desktop, `main_agent`'s MCP wiring, other agent frameworks) can now reach the IRI API without going through LangChain.
- **Skill card** at `src/chemgraph/skills/alcf_iri.md` — first entry in a new `skills/` directory, following the Anthropic Skills convention (YAML frontmatter + when-to-use + auth + example flows + failure modes). Intended for `main_agent`-style skill routing.

Also folds in one small fix carried over from #173 review:

- **Variant-aware auth reprompts** in `alcf_iri_core.py`: `_start_reauth` and the 401 recovery hint now emit tool names that match whichever wrapper module is on the call stack (flat vs category). Was producing category-flavoured instructions ("invoke alcf_auth with action='complete_reauth'") under the flat variant, which caused gpt-4o to hallucinate "session expired" instead of calling `alcf_auth_complete_reauth(auth_code=...)`.

## Files

| Kind | Path |
|---|---|
| MCP server | `src/chemgraph/mcp/alcf_iri_mcp.py` (new) |
| Skill card | `src/chemgraph/skills/alcf_iri.md` (new dir + file) |
| Fix | `src/chemgraph/tools/alcf_iri_core.py` |
| Docs | `docs/mcp_servers.md`, `examples/iri/README.md` |

## Test plan

- [x] `python -c "from chemgraph.mcp import alcf_iri_mcp; ..."` — server module imports cleanly, registers 43 tools with the expected names (`alcf_facility_get`, `alcf_compute_submit_job`, `alcf_auth_start_reauth`, `alcf_auth_complete_reauth`, ...).
- [x] Live call to public `/status/resources` endpoint via `mcp.call_tool('alcf_status_list_resources', {})` returns the same payload as the LangChain flat tools and category tools (HTTP 200, 9 resources, 4 up: Sophia, Crux, Aurora, Polaris).
- [x] Per-tool JSON schemas match `params_schema` in `alcf_iri_core.CATEGORIES` — required fields marked required, optional fields typed `Optional[T]` with `default: null`.
- [x] Variant detection unit-tested against synthetic stack frames simulating both wrapper modules; returns correct value and falls back to `category` when neither module is present.
- [ ] End-to-end MCP handshake via `main_agent` — deferred until main_agent's MCP wiring for IRI lands; nothing here should require changes there.

## Open questions for review

- The `skills/` directory format is a first pass following the Anthropic Skills convention. Happy to rename frontmatter fields (`mcp_server`, `tool_prefix`) or restructure entirely to whatever `main_agent` ends up expecting.
- MCP server defaults to port 9010 for `streamable_http`. If there's an existing port-allocation table for ChemGraph MCP servers, easy to move.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


src/chemgraph/skills/alcf_iri.md
tdpham2, line 1: Could you make an additional skill.md file that is aimed for a coding agent to launch bash command (curl) to use the IRI, not describing the MCP.

Please test it with Claude code to make sure it works well based on the benchmark you created earlier. For the benchmark scoring, make it score either 0 or 1, so that the style doesn't matter, only final answer matters.


I will try to give you a deepagent graph for you to test (similar to Claude code) for you to test next week with the skills.

tdpham2: @JinchuLi2002 Just want to follow up if you have the skills ready. Thanks!

PR #202, opened 2026-08-20
prompt: force executor delegation and preserve reported units

keceli:
Two related failure modes in the multi-agent prompts:

1. The planner would sometimes answer computational chemistry questions from its own knowledge instead of dispatching to executor agents, since it is confident about common molecules. Add an explicit DELEGATION POLICY stating that executor results are the only source of truth, including the "even if you believe you know the answer" case that drives the shortcut.

2. Agents would silently convert units (eV <-> kJ/mol, etc.) while summarizing or aggregating, making values hard to trace back to the tool output that produced them. Add a units rule at every stage that touches a value: - executor: report tool output in original units/precision - planner AGGREGATION (FINISH): no conversion when composing the final answer - aggregator (legacy alias): same rule - formatter: copy value and unit verbatim into the response schema

The conversion rule allows two escapes: an explicit user request, or combining values with no common unit, in which case the conversion must be stated.

The units rule and delegation policy are placed outside the blocks that get_planner_prompt(human_supervised=False) strips, so both survive in the unsupervised prompt.

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Docs
- [ ] Chore / refactor / CI

## Checklist

- [ ] `ruff check .` passes
- [ ] Added/updated tests for the change

### DIFF
--- a/src/chemgraph/prompt/multi_agent_prompt.py
+++ b/src/chemgraph/prompt/multi_agent_prompt.py
@@ -9,6 +9,9 @@ You are an expert in computational chemistry and the **Planner** responsible for
 
 Your role is to act as a router that decomposes user queries into independent subtasks, dispatches them to executor agents, and decides when the workflow is complete.
 
+### DELEGATION POLICY (critical):
+You do NOT execute tools yourself. For any question involving computed properties, molecular data, or simulation results, you MUST dispatch tasks to executor agents and base your answer on their results — even if you believe you already know the answer. Your internal chemistry knowledge may be outdated or wrong; executor results are the only source of truth. Never answer computational questions directly from memory, and never fabricate or anticipate an executor's result.
+
 ### STATE TRANSITION RULES:
 
 **PHASE 1: Task Decomposition (First invocation)**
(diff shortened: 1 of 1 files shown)


src/chemgraph/prompt/multi_agent_prompt.py
```diff
@@ -135,19 +140,24 @@ def get_planner_prompt(human_supervised: bool = True) -> str:
 
 # Legacy alias — the aggregator role is now handled by the planner on FINISH.
 aggregator_prompt = """
```
tdpham2, line 142: The aggregator_prompt is actually not used in the multi-agent graph anymore (and should have been removed) after the last update aimed to fix the graph. You can see the architecture here:

https://github.com/argonne-lcf/ChemGraph/blob/b2db4f6648881f5c064aae8fffb950e605e12fed/src/chemgraph/graphs/multi_agent.py

The unit change is only needed in the planner prompt.

src/chemgraph/prompt/multi_agent_prompt.py
```diff
@@ -9,6 +9,9 @@
 
 Your role is to act as a router that decomposes user queries into independent subtasks, dispatches them to executor agents, and decides when the workflow is complete.
 
+### DELEGATION POLICY (critical):
```
tdpham2, line 12: We discussed about this a long time ago, and we agreed let the planner_agent to have some freedom in deciding whether it should call the executor or not. For example, it should answer "easy" question without having to invoke the executor. 

I think this will help improve the benchmark performance - but just so you understand why I didn't harden the prompt here. We can also enforce the planner -> executor directly in the graph by making their connection a direct edge instead of conditional edge.

keceli: I think it should stay as conditional since some questions might not be answered by the executors when we don't have the relevant tool. But when there is a tool, it should go for it.

keceli: Single agent has "Base all responses strictly on actual tool outputs—never fabricate results, coordinates or SMILES string." I think that prevented it from made up SMILES.

PR #204, opened 2026-08-21
Refactor model loading into protocol × endpoint layers (#201, PR 1)

tdpham2:
## Summary

First of three PRs for #201. Introduces a **protocol × endpoint** architecture for model loading so every caller can eventually follow one pipeline:

```
caller → load_chat_model → endpoint spec → protocol builder → client
```

This PR builds the new layer and proves behavior parity. The central loader and agent callers are migrated in PR 2; config/auth cleanup lands in PR 3.

## What changed

**New `models/protocols/`** — the only construction sites for LangChain clients:
- `openai_compatible.py` (sole `ChatOpenAI`), `anthropic_native.py` (sole `ChatAnthropic`), `google_native.py` (sole `ChatGoogleGenerativeAI`).

**New `models/endpoints/`**:
- `base.py` — internal types `ModelRequest`, `CredentialPolicy`, `EndpointSpec`, `PreparedModel` + shared `resolve_api_key` / `is_local_http_endpoint`.
- One spec per endpoint: `openai_direct`, `argo` (hosted + shim), `alcf` (sophia/minerva/metis), `openrouter`, `vllm`, `anthropic_direct`, `google_direct`. Logic **moved verbatim** from the previous provider loaders.

**Root modules → deprecated shims** (`openai.py`, `anthropic.py`, `gemini.py`, `alcf_endpoints.py`, `openrouter.py`): identical signatures, delegate to the endpoints, emit `DeprecationWarning`, and keep re-exporting their helpers for one release.

## Scope / non-goals for this PR
- **No caller behavior changes.** `loader.py`, `ChemGraph.__init__`, and `run_turn` are untouched here; the two remaining `ChatOpenAI(` sites in `agent/` are migrated in PR 2.
- Academy is out of scope (it inherits the loader via `LLMSettings`).

## Testing
- `ruff check` clean on all new/changed files.
- New `tests/test_endpoints_route_matrix.py` (21 cases across every endpoint) and `tests/test_protocol_construction_guard.py` (no direct construction under `models/` outside `protocols/`).
- Parity: existing `test_openai_model_normalization.py`, `test_alcf_base_url.py`, `test_openrouter_provider.py` all still pass through the shims. 128 model/agent tests green.
- Real-LLM smoke (optional): `pytest tests/test_run_agent.py --run-llm` with a live key exercises shims → endpoints → protocol builders.

Issue #201, opened 2026-08-20
[REFACTOR] Reorganize `src/chemgraph/models/` around protocol × endpoint


InkedWings:
## Problem

The modules under `models/` look organized by API protocol, but the layout
mixes axes: `openai.py` / `anthropic.py` / `gemini.py` are protocol-shaped,
`alcf_endpoints.py` is gateway-shaped, `groq.py` is vendor-shaped, `codex.py`
is product-shaped.

Two consequences:

- **`ChatOpenAI` is built in five places**, two outside `models/`:
  `openai.py:300`/`:317`, `alcf_endpoints.py:120`, `agent/turn.py:294`,
  `agent/llm_agent.py:350`.
- **Dispatch exists twice** — the same `if/elif` chain in
  `models/loader.py:77-113` and `agent/llm_agent.py:298-336`.

The root cause is that two independent dimensions — which protocol to speak,
and which endpoint to speak to — are collapsed into one model-name prefix.

## Proposal

```
models/
  protocols/    openai_compatible.py  anthropic_native.py  google_native.py
  endpoints/    argo.py  alcf.py  openrouter.py  vllm.py
                openai_direct.py  anthropic_direct.py  google_direct.py
  groq.py  local_model.py  codex.py     # own clients, unchanged
  loader.py     # the only dispatch site
```

Endpoint specs declare only the differences — base URL, auth style,
model-name transform, per-model quirks, default params — and reuse the protocol
layer to build the client. One endpoint can then hold several specs (Argo two, ALCF three) and one protocol can back several endpoints.

Worth evaluating whether [LLM Rosetta](https://github.com/oaklight/llm-rosetta) can provide the protocol layer here rather
than us maintaining our own wrappers.

keceli:
I think making use of LLM Rosetta is a good solution. It supports Argo and ALCF inference endpoints as well.

tdpham2:
We agreed to not use LLM Rosetta to keep our implementation simple. All protocol wrappers are provided via langchain. Can circle back to this later if LLM Rosetta provides additional benefits.

tdpham2:
Incorporated in latest release.

PR #205, opened 2026-08-21
Single loader + agent migration (#201, PR 2)

tdpham2:
## Summary

Second of three PRs for #201. This makes load_chat_model the single model-dispatch site and routes the agent layer through it, so agent code no longer constructs LangChain clients directly.

Builds on PR #204, which introduced the protocol × endpoint foundation.

    caller → load_chat_model → endpoint spec → protocol builder → client

### Central loader

- Rebuilt models/loader.py around an ordered EndpointSpec registry:
  codex → openrouter → Argo Anthropic → Argo OpenAI → ALCF → OpenAI → Anthropic → Gemini → Ollama → groq.
- Prefix routes run before catalog checks, preventing prefixed models from being misclassified.
- vLLM/custom routing is used only as the final fallback when a custom base URL is available. Errors from recognized providers propagate instead of falling through.
- Added load_chat_model_prepared to return the client with PreparedModel metadata.
- Added native protocol builders and endpoint specs for Codex, Groq, and Ollama.

### Provider routing fixes

- Reserved the alcf: namespace before the vLLM fallback. Invalid ALCF names now fail in the ALCF endpoint before credential resolution, preventing provider credentials from crossing into a custom endpoint.
- Split Argo into protocol-specific routes:
  - argo:claude-* uses ChatAnthropic and the Anthropic Messages endpoint at /v1/messages.
  - All other argo:* models use ChatOpenAI and the OpenAI-compatible chat-completions endpoint.
- Added the hosted Argo wire mappings for Claude Opus 5, 4.8, and 4.7.
- Normalized existing Argo /v1 and legacy resource-chat URLs to the Anthropic API root for Claude models.
- Kept the deprecated load_openai_model shim returning ChatOpenAI for backward compatibility.

### Agent migration

- Replaced provider-specific construction in ChemGraph.__init__ with one load_chat_model_prepared call.
- Read structured-output support from PreparedModel instead of re-deriving provider behavior in the agent.
- Forward explicit OpenAI API keys through the loader; the previous custom-endpoint branch dropped them.
- Removed the direct ChatOpenAI fallback from turn.py. run_turn and the judge now share the loader seam.
- Retained the existing Codex workflow restriction and caller-side reasoning-effort behavior.

## Verification

- Lint check passes in CI.
- 51 focused loader, endpoint-route, normalization, and client-construction tests pass.
- The construction guard confirms ChatOpenAI, ChatAnthropic, and ChatGoogleGenerativeAI are constructed only under models/protocols.
- Added regressions for ordered routing, provider-error boundaries, ALCF namespace reservation, Argo protocol selection and URL normalization, credential isolation, explicit-key forwarding, and shared agent loader seams.

## Scope / non-goals

- PR3 will handle configuration and authentication cleanup, including dedicated Argo and vLLM settings.
- PR3 will centralize reasoning-effort validation/defaulting so PreparedModel reports the effective value consistently.
- Academy remains out of scope and continues to inherit model loading through LLMSettings.

### DIFF
--- a/src/chemgraph/models/supported_models.py
+++ b/src/chemgraph/models/supported_models.py
@@ -150,12 +150,16 @@ supported_openrouter_models = [
     "openrouter:minimax/minimax-m2.7",
 ]
 
-# Default Argo API base URL (used when no --base-url is provided).
+# Default Argo OpenAI-compatible API base URL.
 ARGO_DEFAULT_BASE_URL = "https://apps.inside.anl.gov/argoapi/v1"
 
+# Default Argo Anthropic-native API base URL. ChatAnthropic appends
+# ``/v1/messages`` to this root.
+ARGO_DEFAULT_ANTHROPIC_BASE_URL = "https://apps.inside.anl.gov/argoapi"
+
 # Argo models -- all use the "argo:" prefix.
-# Which endpoint they hit depends on --base-url / config.
-# Default: ARGO_DEFAULT_BASE_URL (Argo API).
+# Claude models use the Anthropic-native protocol; all others use the
+# OpenAI-compatible protocol. ``--base-url`` / config may override either URL.
 supported_argo_models = [
     # GPT family
     "argo:gpt-4o",
@@ -197,7 +201,7 @@ supported_argo_models = [
     "argo:claude-sonnet-4.5",
 ]
 
-# Exact Argo model routes that require minimal ChatOpenAI construction.
+# Exact Argo model routes that require minimal client construction.
 # Optional sampling parameters are omitted for every entry in this set. Add
 # entries only after validating them against the deployed endpoint.
 MODELS_WITHOUT_TEMPERATURE = frozenset(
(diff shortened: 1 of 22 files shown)


src/chemgraph/models/supported_models.py
```diff
@@ -197,7 +201,7 @@
     "argo:claude-sonnet-4.5",
 ]
 
-# Exact Argo model routes that require minimal ChatOpenAI construction.
+# Exact Argo model routes that require minimal client construction.
 # Optional sampling parameters are omitted for every entry in this set. Add
 # entries only after validating them against the deployed endpoint.
 MODELS_WITHOUT_TEMPERATURE = frozenset(
```
InkedWings, line 207: Need to add opus 4.7, 4.8, 5 into the list.

tdpham2: Thanks! Added.

tdpham2:
Thanks @InkedWings for reviewing this PR.

Issue #194, opened 2026-08-19
[FEATURE] LLM Local Inference on ALCF Aurora


williamyxl:
Proposing a feature with existing solution: ChemGraph operation need the LLM inference service. What if we run inference locally on Aurora compute nodes. 

I have some recent recipes for running LLM inference on Aurora compute node for: 
1. GPT-OSS-120B: single serve ~40 tok/s, high concurrency: ~4500 tok/s; (https://github.com/williamyxl/alcf-aurora-llm/blob/main/gpt-oss-120b/build-vllm-xpu/opus4.8-new_campaign/BEST_RECIPES.md)
2. Inkling-975B: single serve ~6 tok/s; (https://github.com/williamyxl/alcf-aurora-llm/blob/main/inkling/build-llamacpp-sycl/BEST_RECIPE.md)

This should eliminate the need for data connection to other inference API services inside or outside Argonne. 

If more models are needed and comply with ALCF/Argonne policy (open weight and made in US), I can add them. 

Right now, I'm working on Nvidia's Nemotron 3 Ultra inference. Let me know your opinions, please. 

Thanks.
-Xiaoli

tdpham2:
Thanks @williamyxl. These recipes look very useful. I'm just not sure yet where this should fit within ChemGraph.

Currently, ChemGraph connects to an LLM through `model_name` and `base_url`, with the assumption that the inference endpoint is already available (cloud provider, ALCF inference service). So actually launching and managing an inference server on Aurora is currently outside ChemGraph's scope.

I see two possible directions:

1. Have a separate deployment repo, similar to MOFA, that sets up both ChemGraph and the vLLM inference endpoints on Aurora for a specific application. This would keep the application/HPC deployment logic outside the main ChemGraph repo.
2. Keep the Aurora inference setup as a separate deployment recipe that only exposes an OpenAI-compatible endpoint, while ChemGraph continues to connect through `model_name` and `base_url` without needing to know how or where the model is deployed. We can test this deployment with ChemGraph and highlight it in ChemGraph's documentation.

williamyxl:
I have made some progress. Just checking, do I have permission to push a new branch and open PR in this repo?

williamyxl:
created PR: https://github.com/argonne-lcf/ChemGraph/pull/203

PR #207, opened 2026-08-21
Add first-class Aurora on-node inference provider (aurora:)

williamyxl:
## Summary

Adds a first-class `aurora:` prefix provider for OpenAI-compatible LLM servers
running on ALCF Aurora compute nodes (llama.cpp SYCL `llama-server` or
vLLM-XPU), analogous to the existing `alcf:` provider.

- **`models/aurora_endpoints.py`**: `AURORA_MODEL_PREFIX` and `load_aurora_model`.
  Base URL is resolved from `--base-url` → `AURORA_BASE_URL` env →
  `[api.aurora].base_url` in config → `http://127.0.0.1:8000/v1` default. API key
  defaults to `"dummy"` (on-node servers don't enforce auth).
- **Dispatch wiring**: `models/loader.py` and `agent/llm_agent.py` route
  `aurora:` models to `load_aurora_model`.
- **Config resolution**: `utils/config_utils.py` (flat + nested config) and
  `models/settings.py` (CLI TOML extraction) honor `[api.aurora].base_url`.
- **`models/supported_models.py`**: `AURORA_DEFAULT_BASE_URL` and
  `supported_aurora_models` (`aurora:gpt-oss-120b`, `aurora:inkling`,
  `aurora:nemotron-3-ultra`, `aurora:nemotron-4-340b`). The list is for
  *discovery* only (`chemgraph models` / dropdown); dispatch is by prefix, so any
  served model id also works.
- **`config.toml`**: `[api.aurora]` section.
- **`docs/running_local_models.md`**: Aurora provider usage section.

The served model id (after the `aurora:` prefix is stripped) must match the
server's advertised name (`llama-server --alias` / vLLM `--served-model-name`)
and must support OpenAI tool calling.

## Related issues

Part of #194.

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Docs
- [ ] Chore / refactor / CI

## How was this tested?

- 13 hermetic tests in `tests/test_aurora_endpoints.py` covering prefix
  stripping, base URL resolution (flat + nested config, env var override, default
  fallback), `load_aurora_model` construction, `load_chat_model` dispatch, and
  `LLMSettings` TOML extraction. No network required.
- `ruff check .` and `pytest tests/ -k "not tblite"` pass.
- End-to-end on Aurora: EMT geometry optimization via the `aurora:` provider
  (`smiles_to_coordinate_file` → `run_ase` → converged, energy reported) with
  `overall_rc=0`, validated against locally-served gpt-oss-120b, Inkling,
  Nemotron-3-Ultra, and Nemotron-4-340B.

## Checklist

- [x] `ruff check .` passes

### DIFF
--- a/config.toml
+++ b/config.toml
@@ -64,6 +64,15 @@ timeout = 30
 base_url = "https://inference-api.alcf.anl.gov/resource_server/sophia/vllm/v1"
 timeout = 30
 
+# Aurora on-node inference (llama.cpp SYCL llama-server or vLLM-XPU), served as
+# an OpenAI-compatible /v1 endpoint. Use "aurora:<served-model-id>" models. The
+# node IP changes per job and compute nodes have no public IP, so set base_url
+# to a co-located server (127.0.0.1) or an SSH tunnel from a login node. May
+# also be set via the AURORA_BASE_URL environment variable.
+[api.aurora]
+base_url = "http://127.0.0.1:8000/v1"
+timeout = 60
+
 [api.local]
 base_url = "http://localhost:11434"
 timeout = 60
--- a/docs/running_local_models.md
+++ b/docs/running_local_models.md
@@ -45,6 +45,34 @@ route through a compatible ChemGraph provider/model ID, and implement reliable
 tool calling. Review the script's hardware/model assumptions and test a
 non-destructive query first. Ollama remains the documented first local route.
 
+## Aurora on-node inference (`aurora:` models)
+
+ChemGraph has a first-class provider for on-node LLM servers on ALCF Aurora
+(llama.cpp SYCL `llama-server` or vLLM-XPU) that expose an OpenAI-compatible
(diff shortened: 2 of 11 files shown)


tdpham2:
Thanks for putting this together and for testing it on Aurora. I don’t think we need a separate aurora: provider in ChemGraph. We generally reserve provider-specific routes for hosted inference services (openai, argo, alcf inference service) where ChemGraph can simplify the user setup or handle provider-specific behavior. In this case, users still need to launch and manage the inference server themselves, and once it is running, it exposes the same OpenAI-compatible interface already supported by the vLLM/custom endpoint.

Could you rework this as documentation (a specific example) showing how to use the Aurora recipes with --base-url, VLLM_BASE_URL, or [api.vllm].base_url, along with the server’s actual model ID? If the current vllm setup doesn't work, please help us improve it instead of adding a separate aurora: endpoint.

This would preserve the useful Aurora guidance for others while avoiding another configuration.

PR #219, opened 2026-08-26
Add reusable Deep Agent workflow

tdpham2:
## Summary

Add a reusable workspace Deep Agent that runs through `ChemGraph(workflow_type="deep_agent")` or as the `deepagent` worker under `main_agent`.

- Share graph construction, skills, workspace tools, and structured action approvals across Python and CLI entry points.
- Expose host files through one `/workspace` namespace while preserving shell execution and host-path guidance.
- Support repeated approval rounds, concurrent interrupt recovery, asynchronous checkpoints, and thread-specific message persistence.
- Preserve inactive configuration for interactive workflow switches and report invalid CLI input cleanly.
- Document workspace confirmation, shell scope, explicit command environments, registry names, and asynchronous checkpointer usage.

Rebased onto current `main`, preserving #232's session-history fix. Review fixes retain the existing confirmed current-directory default for interactive use; headless runs require an explicit workspace and approval-skip flag.

## Related issues

None.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Docs
- [ ] Chore / refactor / CI

## How was this tested?

- `ruff check .`
- `pytest tests/ -k "not tblite"`: **1,122 passed, 21 skipped, 2 deselected**
- Real graph regressions cover twelve approved file writes through Python and CLI, concurrent interrupts, unique sync/async workspace searches, registry delegation, and async SQLite approval recovery.
- Configuration and persistence regressions cover workflow precedence, inactive settings, malformed input, resumed transcripts, and idless messages across threads.
- Tests use the rebased runtime pins and an isolated session database. No live-LLM, live-Globus, or live-EnsembleLauncher flags were enabled.

## Checklist

- [x] Rebased onto current `main`
- [x] Preserved the original PR's feature commits and first-round review fixes
- [x] Verified later review findings and addressed substantiated issues
- [x] Ruff and the full local test suite pass
- [x] Added regression coverage and updated user-facing documentation

### DIFF
--- a/docs/python_api.md
+++ b/docs/python_api.md
@@ -89,6 +89,10 @@ agent = ChemGraph(
         root_dir="/path/to/checkout",
         virtual_mode=True,
     ),
+    deepagent_skills=[
+        "/workspace/shared-skills/",
+        "/workspace/.agents/skills/",
+    ],
 )
 result = await agent.run(
     "Review the test failures.",
@@ -102,10 +106,21 @@ tool path `/workspace/src/example.py` maps directly to
 also supplies that host path for shell commands. Custom backends, existing
 composite backends, and non-virtual local backends are passed through unchanged.
 
+`deepagent_skills` contains ordered POSIX paths interpreted by that same
+backend. ChemGraph passes them to Deep Agents without scanning any implicit
+user or project locations. Each source contains skill directories with a
+required `SKILL.md`; later sources override earlier sources with the same
+skill name. With `StateBackend`, callers invoking the graph directly must seed
+the corresponding files in graph state before the skills can load.
+
 The default approval policy interrupts before shell commands and file
 mutations. Without a `human_input_handler`, `run()` raises
 `HumanInputRequired`; its `payload` retains the structured Deep Agents action
-requests and must be resumed with matching structured decisions. Setting
+requests and must be resumed with matching structured decisions. Its
+`interrupts` tuple retains every pending request and its LangGraph interrupt
+ID; callers must resume concurrent requests with an exact mapping from those
+IDs to responses. `question` and `payload` continue to describe the first
+request for compatibility. Setting
 `deepagent_auto_approve=True` removes this boundary and should be limited to an
 externally isolated, explicitly trusted workspace.
 
@@ -116,7 +131,12 @@ from chemgraph.graphs.deep_agent import construct_deep_agent_graph
 from chemgraph.registry import AgentRegistry
 
 standalone_graph = construct_deep_agent_graph(model, backend=backend)
-worker = AgentRegistry().as_subagent("deepagent", llm=model, backend=backend)
+worker = AgentRegistry().as_subagent(
+    "deepagent",
+    llm=model,
+    backend=backend,
+    skills=["/workspace/.agents/skills/"],
+)
 ```
 
 `as_subagent()` compiles the graph with `checkpointer=None` so it inherits its
--- a/docs/python_api.md
+++ b/docs/python_api.md
@@ -120,7 +120,10 @@ requests and must be resumed with matching structured decisions. Its
 `interrupts` tuple retains every pending request and its LangGraph interrupt
 ID; callers must resume concurrent requests with an exact mapping from those
 IDs to responses. `question` and `payload` continue to describe the first
-request for compatibility. Setting
+request for compatibility. A configured handler may use the legacy
+`handler(question)` signature or `handler(question, payload)` when it needs the
+raw structured request; both synchronous and asynchronous handlers are
+supported. Setting
 `deepagent_auto_approve=True` removes this boundary and should be limited to an
 externally isolated, explicitly trusted workspace.
 
--- a/docs/python_api.md
+++ b/docs/python_api.md
@@ -73,6 +73,56 @@ For CLI use, the equivalent is:
 chemgraph run --interactive --workflow main_agent
 ```
 
+## Workspace Deep Agent
+
+The workspace workflow is separately callable through `ChemGraph.run()`:
+
+```python
+from deepagents.backends import LocalShellBackend
+
+from chemgraph.agent.llm_agent import ChemGraph
+
+agent = ChemGraph(
+    model_name="claude-sonnet-4-20250514",
+    workflow_type="deep_agent",
+    deepagent_backend=LocalShellBackend(
+        root_dir="/path/to/checkout",
+        virtual_mode=True,
+    ),
+)
+result = await agent.run(
+    "Review the test failures.",
+    config={"configurable": {"thread_id": "workspace-review"}},
+)
+```
+
+A virtual `LocalShellBackend` is exposed to the agent at `/workspace`, so file
+tool path `/workspace/src/example.py` maps directly to
+`/path/to/checkout/src/example.py`. The generated Deep Agents system context
+also supplies that host path for shell commands. Custom backends, existing
+composite backends, and non-virtual local backends are passed through unchanged.
+
+The default approval policy interrupts before shell commands and file
+mutations. Without a `human_input_handler`, `run()` raises
+`HumanInputRequired`; its `payload` retains the structured Deep Agents action
+requests and must be resumed with matching structured decisions. Setting
+`deepagent_auto_approve=True` removes this boundary and should be limited to an
+externally isolated, explicitly trusted workspace.
+
+The same constructor can be composed as a worker:
+
+```python
+from chemgraph.graphs.deep_agent import construct_deep_agent_graph
+from chemgraph.registry import AgentRegistry
+
+standalone_graph = construct_deep_agent_graph(model, backend=backend)
+worker = AgentRegistry().as_subagent("deepagent", llm=model, backend=backend)
+```
+
+`as_subagent()` compiles the graph with `checkpointer=None` so it inherits its
+parent checkpoint. `construct_main_agent_graph(enable_deepagent=True, ...)`
+uses this same workflow under the stable subagent name `deepagent`.
+
 ## Custom tools
 
 `ChemGraph` can be extended with compatible LangChain tools. Keep tools narrow,
--- a/docs/python_api.md
+++ b/docs/python_api.md
@@ -78,6 +78,8 @@ chemgraph run --interactive --workflow main_agent
 The workspace workflow is separately callable through `ChemGraph.run()`:
 
 ```python
+import os
+
 from deepagents.backends import LocalShellBackend
 
 from chemgraph.agent.llm_agent import ChemGraph
@@ -88,6 +90,12 @@ agent = ChemGraph(
     deepagent_backend=LocalShellBackend(
         root_dir="/path/to/checkout",
         virtual_mode=True,
+        env={
+            name: os.environ[name]
+            for name in ("PATH", "PYTHONPATH", "VIRTUAL_ENV", "CONDA_PREFIX", "TMPDIR")
+            if name in os.environ
+        },
+        inherit_env=False,
     ),
     deepagent_skills=[
         "/workspace/shared-skills/",
@@ -143,8 +151,15 @@ worker = AgentRegistry().as_subagent(
 ```
 
 `as_subagent()` compiles the graph with `checkpointer=None` so it inherits its
-parent checkpoint. `construct_main_agent_graph(enable_deepagent=True, ...)`
+parent checkpoint. The registry returns the canonical worker name `deep_agent`,
+even when requested through the `deepagent` alias; use `worker["name"]` when
+composing task calls. `construct_main_agent_graph(enable_deepagent=True, ...)`
 uses this same workflow under the stable subagent name `deepagent`.
+
+Caller-owned asynchronous checkpointers, including `AsyncSqliteSaver`, are
+supported by `await agent.run(...)` and `await agent.apersist_run_state(config)`.
+Keep the saver open on the same event loop for the run and any resumes. The
+synchronous state methods remain available for synchronous checkpointers.
 
 ## Custom tools
 
--- a/src/chemgraph/agent/interrupts.py
+++ b/src/chemgraph/agent/interrupts.py
@@ -1,0 +1,65 @@
+"""Shared normalization helpers for resumable LangGraph interrupts."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass(frozen=True)
+class PendingInterrupt:
+    """One pending request for user input."""
+
+    id: str
+    payload: Any
+
+
+def normalize_interrupts(values: Any) -> list[PendingInterrupt]:
+    """Convert LangGraph interrupt values into stable application records."""
+    if values is None:
+        return []
+    if not isinstance(values, (list, tuple)):
+        values = [values]
+
+    return [
+        PendingInterrupt(
+            id=str(getattr(item, "id", "") or ""),
+            payload=getattr(item, "value", item),
+        )
+        for item in values
+    ]
+
+
+def deduplicate_interrupts(
+    interrupts: list[PendingInterrupt],
+) -> tuple[PendingInterrupt, ...]:
+    """Deduplicate streamed and checkpointed copies of pending interrupts."""
+    unique: list[PendingInterrupt] = []
+    seen: set[str] = set()
+    for item in interrupts:
+        key = item.id or repr(item.payload)
+        if key in seen:
+            continue
+        seen.add(key)
+        unique.append(item)
+    return tuple(unique)
+
+
+def interrupt_question(payload: Any) -> str:
+    """Extract readable question text from an interrupt payload."""
+    if isinstance(payload, dict):
+        return str(
+            payload.get(
+                "question",
+                payload.get("message", payload.get("instruction", payload)),
+            )
+        )
+    return str(payload)
+
+
+__all__ = [
+    "PendingInterrupt",
+    "deduplicate_interrupts",
+    "interrupt_question",
+    "normalize_interrupts",
+]
--- a/src/chemgraph/agent/interrupts.py
+++ b/src/chemgraph/agent/interrupts.py
@@ -23,7 +23,11 @@ def normalize_interrupts(values: Any) -> list[PendingInterrupt]:
 
     return [
         PendingInterrupt(
-            id=str(getattr(item, "id", "") or ""),
+            id=(
+                str(getattr(item, "id", "") or "")
+                if getattr(item, "id", "") != "placeholder-id"
+                else ""
+            ),
             payload=getattr(item, "value", item),
         )
         for item in values
@@ -37,12 +41,39 @@ def deduplicate_interrupts(
     unique: list[PendingInterrupt] = []
     seen: set[str] = set()
     for item in interrupts:
-        key = item.id or repr(item.payload)
-        if key in seen:
+        if item.id and item.id in seen:
             continue
-        seen.add(key)
+        if item.id:
+            seen.add(item.id)
         unique.append(item)
     return tuple(unique)
+
+
+def collect_pending_interrupts(
+    streamed: list[PendingInterrupt], snapshot: Any = None, *, fallback: bool = False,
+) -> tuple[PendingInterrupt, ...]:
+    """Prefer checkpoint requests over streamed copies of the same pause.
+
+    Anonymous requests cannot be matched by payload: distinct actions may have
+    identical payloads. Use one authoritative collection instead.
+    """
+    checkpointed = normalize_interrupts(getattr(snapshot, "interrupts", ()))
+    if not checkpointed:
+        for task in getattr(snapshot, "tasks", ()):
+            checkpointed.extend(normalize_interrupts(getattr(task, "interrupts", ())))
+    pending = deduplicate_interrupts(checkpointed or streamed)
+    if fallback and not pending:
+        return (PendingInterrupt(id="", payload={"question": "The workflow needs your input."}),)
+    return pending
+
+
+def is_tool_review(payload: Any) -> bool:
+    """Return whether an interrupt contains structured action reviews."""
+    return (
+        isinstance(payload, dict)
+        and isinstance(payload.get("action_requests"), list)
+        and isinstance(payload.get("review_configs"), list)
+    )
 
 
 def interrupt_question(payload: Any) -> str:
@@ -59,7 +90,9 @@ def interrupt_question(payload: Any) -> str:
 
 __all__ = [
     "PendingInterrupt",
+    "collect_pending_interrupts",
     "deduplicate_interrupts",
     "interrupt_question",
+    "is_tool_review",
     "normalize_interrupts",
 ]
--- a/src/chemgraph/agent/llm_agent.py
+++ b/src/chemgraph/agent/llm_agent.py
@@ -12,7 +12,8 @@ import uuid
 from chemgraph.agent.events import EventCallback, _AstreamEventCallback
 from chemgraph.agent.interrupts import (
     PendingInterrupt,
-    deduplicate_interrupts,
+    collect_pending_interrupts,
+    is_tool_review,
     interrupt_question,
     normalize_interrupts,
 )
@@ -256,7 +257,6 @@ class ChemGraph:
         reasoning_effort: Optional[str] = None,
         checkpointer: BaseCheckpointSaver | None = None,
     ):
-        normalized_deepagent_skills = normalize_skill_sources(deepagent_skills)
         if enable_deepagent and workflow_type != "main_agent":
             raise ValueError(
                 "enable_deepagent is supported only for the main_agent workflow."
@@ -268,7 +268,7 @@ class ChemGraph:
                 "deepagent_backend requires enable_deepagent=True or "
                 "workflow_type='deep_agent'."
             )
-        if normalized_deepagent_skills and not (
+        if deepagent_skills and not (
             enable_deepagent or workflow_type == "deep_agent"
         ):
             raise ValueError(
@@ -297,6 +297,7 @@ class ChemGraph:
                 "Experimental codex: models currently support only the "
                 "single_agent, main_agent, and deep_agent workflows."
             )
+        normalized_deepagent_skills = normalize_skill_sources(deepagent_skills)
         reasoning_effort = _resolve_reasoning_effort(model_name, reasoning_effort)
 
         # Always generate a unique identifier for this instance
@@ -711,6 +712,17 @@ class ChemGraph:
         dict or str
             Dictionary of metadata if successful, or "Error" if failed.
         """
+        if config is None:
+            config = {"configurable": {"thread_id": "1"}}
+        try:
+            state = self.get_state(config=config)
+        except Exception as exc:
+            print("Error with write_state: ", str(exc))
+            return "Error"
+        return self._write_state_snapshot(state, config, file_path, file_name)
+
+    def _write_state_snapshot(self, state, config, file_path=None, file_name=None):
+        """Serialize an already retrieved checkpoint for either runner."""
         import json
         import subprocess
 
@@ -730,7 +742,6 @@ class ChemGraph:
                     file_name = f"state_thread_{thread_id}_{self.uuid}_{timestamp}.json"
                 file_path = os.path.join(log_dir, file_name)
 
-            state = self.get_state(config=config)
             serialized_state = serialize_state(state)
 
             try:
@@ -940,6 +951,35 @@ class ChemGraph:
             "Use 'last_message' or 'state'."
         )
 
+    async def apersist_run_state(self, config: dict) -> dict | None:
+        """Retrieve and log a checkpoint without calling a sync saver API."""
+        try:
+            state = await self.aget_state(config)
+        except Exception:
+            logger.warning("Could not log workflow checkpoint.", exc_info=True)
+            return None
+        self._write_state_snapshot(state, config)
+        return state
+
+    async def afinalize_completed_run(
+        self, last_state: dict, config: dict, query: str,
+    ) -> Any:
+        """Persist a completed async run and return the configured result."""
+        self._save_messages_to_store(
+            last_state, query, thread_id=str(config["configurable"]["thread_id"])
+        )
+        state = await self.apersist_run_state(config)
+        if self.return_option == "last_message":
+            return last_state["messages"][-1]
+        if self.return_option == "state":
+            if state is None:
+                state = await self.aget_state(config)
+            return serialize_state(state)
+        raise ValueError(
+            f"Unsupported return_option: {self.return_option}. "
+            "Use 'last_message' or 'state'."
+        )
+
     def load_previous_context(
         self,
         session_id: str,
@@ -1081,6 +1121,7 @@ class ChemGraph:
             prev_msgs: list = []
             last_st = None
             found_interrupts: list[PendingInterrupt] = []
+            bare_interrupt = False
             try:
                 async for s in self.workflow.astream(
                     stream_input, stream_mode="values", config=cfg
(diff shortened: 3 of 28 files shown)


tdpham2:
@JinchuLi2002 Can you test this with iri skills.md?

tdpham2:
Thanks @JinchuLi2002 for checking this with the IRI API

src/chemgraph/agent/llm_agent.py
```diff
@@ -392,11 +446,19 @@ def append_calculator_context(prompt: str) -> str:
             "terminal_tool_names": self.terminal_tool_names,
             "enable_deepagent": self.enable_deepagent,
             "workspace": str(workspace) if workspace is not None else None,
+            "deepagent_skills": (
+                self.deepagent_skills if self.enable_deepagent else ()
+            ),
```
keceli: This invalidates every existing durable `main_agent` checkpoint, including for users who never enable Deep Agent.

The key is added to `topology_payload` unconditionally, so the sha256 below changes for everyone. `package_version` isn't part of the payload, so nothing else would have shifted the fingerprint — on their next resume they'd hit `IncompatibleCheckpointError("The active graph topology does not match the stored session.")` from `main_session.py:173`, which gives no hint that an upgrade caused it.

The block immediately after this one already has the right pattern:

```python
if self.enable_deepagent and self.deepagent_prompt != DEFAULT_DEEPAGENT_PROMPT:
    topology_payload["deepagent_prompt"] = self.deepagent_prompt
```

The same treatment here would keep the fingerprint stable for anyone not using skills:

```python
if self.enable_deepagent and self.deepagent_skills:
    topology_payload["deepagent_skills"] = self.deepagent_skills
```

Otherwise `graph_schema_version` should be bumped, so the failure at least reports itself as a version mismatch.

src/chemgraph/agent/llm_agent.py
```diff
@@ -838,7 +956,11 @@ async def _call_human_input_handler(self, question: str) -> str:
...
         if handler is None:
-            raise HumanInputRequired(question)
+            raise HumanInputRequired(
+                question,
+                payload=payload,
+                interrupts=interrupts,
+            )
         if asyncio.iscoroutinefunction(handler):
             return await handler(question)
         return handler(question)
```
keceli: `payload` and `interrupts` are accepted by `_call_human_input_handler` but only ever used to construct the `HumanInputRequired` exception — the handler itself still receives just `question`.

That leaves the programmatic path unable to do what the new docs ask of it. The class docstring now says Deep Agent action reviews "require a structured decision mapping rather than a string", but for a tool-review payload `interrupt_question()` finds no `question` / `message` / `instruction` key and falls through to `str(payload)`. So a caller passing `human_input_handler=` to `ChemGraph(workflow_type="deep_agent")` has to parse a dict repr back out of a string to decide approve/reject.

The CLI is unaffected — it goes through the `HumanInputRequired` path into `_prompt_for_interrupt`, which gets the real payload. It's only the handler callback that's short of what it needs. Forwarding `payload` to handlers whose signature accepts it would close the gap.

keceli (commented):
Thanks for this — the refactor direction is right. Pulling `construct_deep_agent_graph` out so the standalone workflow and the `main_agent` worker share one prompt, backend, and approval policy removes real duplication, and hoisting interrupt normalization out of `main_session.py`/`llm_agent.py` into `agent/interrupts.py` is a good cleanup on its own.

Two things I'd like addressed before merge (left inline), plus a CI concern.

### CI has not run on the head commit

`da3adfe` has zero check runs. The only runs on this branch are against the first of the three commits, and they came back `startup_failure` / `cancelled` / `failure` — the failures were runner acquisition (`The job was not acquired by Runner of type hosted`), not code, so nothing is known-broken. But "Expose Deep Agent through the CLI" and "Add skills support to the Deep Agent workflow" have never been exercised in CI at all. Could you push a rebase or an empty commit to force a fresh run? For +1879 lines touching the CLI entry points I'd rather not merge on local results alone.

### Smaller points

- **Private names crossing module boundaries.** `_normalize_skill_sources` is imported into both `llm_agent.py` and `cli/commands.py`, and `cli/commands.py` reaches into `agent._finalize_completed_run(...)` and `agent._persist_run_state(...)`. These look like deliberate seams — if so, drop the leading underscore so the contract is explicit.
- **`DEFAULT_DEEPAGENT_PROMPT` depends on a `deepagents`-internal string.** The prompt tells the model to follow the "Shell paths vs. virtual paths" mapping, which the library emits, not us. `deepagents==0.7.5` is pinned exactly and `test_virtual_workspace_supplies_shell_host_path_to_model` asserts the string, so an upgrade fails loudly — that's the right guard. Flagging it as a known blocker for whoever bumps that pin.
- **`--deepagent-workspace` reads as a sandbox in headless mode.** Requiring both an explicit workspace and `--deepagent-dangerously-skip-approvals` is a reasonable gate, but the shell is not confined to that directory. The warning panel says so; the flag's `--help` text doesn't. Worth repeating there, since the mandatory-workspace requirement implies containment that only covers the file tools. `inherit_env=False` plus the env allowlist on the backend is a nice touch.

### Things I liked

- CLI resume now routes through `_finalize_completed_run`, so session-store messages actually get persisted on the resume path — that was silently dropped before.
- `_%f` on the state-log timestamp; the new pending-checkpoint write on `HumanInputRequired` would have collided within the same second otherwise.
- Docs are unusually candid about the risk surface, and the 22 new tests include real graph invocation rather than only mocks.

I couldn't run the suite myself — my checkout's venv is stale against the pinned `langgraph==1.2.9` / `deepagents==0.7.5`.

keceli:
These were claude code (opus 5) comments, by the way. I thought it would be signed by it.

src/chemgraph/agent/llm_agent.py
tdpham2: Fixed in 7933e6285. Existing one-argument handlers remain compatible, while handlers whose signature accepts a second argument now receive the original structured payload. Signature binding happens before invocation, so a `TypeError` raised inside the callback is not mistaken for an arity mismatch. Awaitable return values are also handled, including async callable objects. Tests cover raw payload identity, async dispatch, legacy one-argument handlers, and callback error propagation.

src/chemgraph/agent/llm_agent.py
tdpham2: Fixed in 7933e6285. `deepagent_skills` now enters the topology payload only when Deep Agent is enabled and the ordered skill tuple is non-empty. The field remains persisted in `MainAgentGraphConfig`, and `graph_schema_version` stays at 1 because default/no-skill configurations retain their legacy hash. The regression test reconstructs the pre-skills fingerprint for both disabled and enabled-without-skills configurations and asserts exact equality; configured skill sets still produce distinct fingerprints.

tdpham2:
Addressed the remaining review feedback in 7933e6285 and the Windows CI portability follow-up in b4fc462b7. The cross-module seams now use explicit public names (`normalize_skill_sources`, `persist_run_state`, and `finalize_completed_run`), and the CLI help now states that the workspace scopes file tools but does not confine shell commands. I left the `DEFAULT_DEEPAGENT_PROMPT` wording unchanged: the exact `deepagents==0.7.5` pin plus `test_virtual_workspace_supplies_shell_host_path_to_model` intentionally make that dependency an upgrade gate. The refreshed head is green across lint, Linux/macOS/Windows 3.11–3.13 tests, extras, wheel installs, optional-install checks, and conda-MACE.

src/chemgraph/graphs/deep_agent.py
```diff
@@ -0,0 +1,128 @@
...
+        normalized.append(source)
+    return tuple(normalized)
+
+
+def _normalize_backend(backend: BackendProtocol) -> BackendProtocol:
+    """Mount a virtual local workspace at the path Deep Agent expects."""
+    if isinstance(backend, LocalShellBackend) and backend.virtual_mode:
+        return CompositeBackend(
+            default=backend,
+            routes={_WORKSPACE_MOUNT: backend},
```
keceli, line 66: `default` and the `/workspace/` route are the **same object**, so `CompositeBackend` searches it twice and every file becomes reachable under two names.

Against `deepagents==0.7.5` with a root containing `a.py` and `sub/b.py`, `glob('**/*.py')` returns 4 hits for 2 files (`/a.py`, `/sub/b.py`, `/workspace/a.py`, `/workspace/sub/b.py`), `grep('TODO','/')` likewise doubles, and `ls('/')` shows a phantom `/workspace/` beside the real entries.

Beyond the wasted tokens, the model can edit `/a.py` and `/workspace/a.py` believing they're different files while they silently clobber each other, and duplicate hits burn half the `max_count` budget so results get flagged `truncated=True` early.

This also regresses the existing path: on `main`, `main_agent --deepagent` passed the `LocalShellBackend` straight to `create_deep_agent` with no composite wrapper.

Routing `/workspace/` to the backend while giving `default` something else (or dropping `default` entirely) would give the single namespace the prompt already assumes.

src/chemgraph/agent/llm_agent.py
```diff
@@ -1045,8 +1175,8 @@ async def _stream_until_interrupt(stream_input, cfg):
             # the first answer).
             max_interrupts = 10  # safety guard against infinite interrupt loops
```
keceli, line 1176: This cap makes approval-gated Deep Agent runs unusable for anything real.

`DEFAULT_DEEPAGENT_INTERRUPT_ON` pauses on every `execute` / `write_file` / `edit_file` / `delete`, so the interrupt count tracks *actions taken*, not user sessions. A deep_agent asked to write 12 files raised `RuntimeError: Workflow exceeded maximum of 10 human interrupts` after the user had approved 10 of them — 10 files already on disk, `run()` returned nothing. `run_query` hits the same wall at commands.py:467 with "Exceeded maximum number of human interrupts. Aborting." and returns `None`.

The `+= len(pending_interrupts)` change also means one round surfacing 11 concurrent reviews aborts before the user is prompted even once.

The guard was written for `ask_human`, where 10 questions really is a runaway loop. For a workflow that interrupts on every mutation it needs to either be much larger, count *resume rounds* rather than actions, or not apply to approval-type interrupts at all.

src/chemgraph/cli/main.py
```diff
@@ -512,17 +537,64 @@ def _handle_run(args: argparse.Namespace) -> None:
...
-            console.print("[red]--deepagent-workspace requires --deepagent.[/red]")
+            console.print(
+                "[red]--deepagent-workspace requires --deepagent or "
+                "-w deep_agent.[/red]"
+            )
+            sys.exit(2)
+    if deepagent_skills and not (
+        enable_deepagent or args.workflow == "deep_agent"
+    ):
+        if cli_deepagent is False and cli_deepagent_skills is None:
```
keceli, line 564: This escape hatch never fires for config-file users, so the `deepagent_skills` line this PR adds to `config.toml` breaks **every** run.

`--deepagent` is `BooleanOptionalAction` with `default=None`, so when the user hasn't typed the flag `cli_deepagent` is `None`, not `False`. The condition therefore fails and control falls to the `sys.exit(2)` below.

Reproduced with the shipped `config.toml` — uncomment line 15 (`deepagent_skills = [...]`), leave `enable_deepagent = false` and `workflow = "single_agent"` exactly as shipped, and `chemgraph run -c config.toml -q '...'` dies with "--deepagent-skill requires --deepagent or -w deep_agent." Every invocation with that config file is dead, including ones with nothing to do with Deep Agent.

`cli_deepagent is not True` would match the intent. Worth checking the same pattern on the `deepagent_workspace` branch just above.

Separately: the `config.toml` comment suggesting `workflow = "deep_agent"` as the alternative is inert, since `--workflow` has a non-`None` argparse default and so is never overwritten by the merge loop.

src/chemgraph/cli/main.py
```diff
@@ -512,17 +537,64 @@ def _handle_run(args: argparse.Namespace) -> None:
...
     enable_deepagent = bool(getattr(args, "deepagent", False))
     deepagent_workspace = getattr(args, "deepagent_workspace", None)
-    if deepagent_workspace is not None and not enable_deepagent:
+    deepagent_skills = getattr(args, "deepagent_skills", None)
+    deepagent_auto_approve = bool(
+        getattr(args, "deepagent_dangerously_skip_approvals", False)
+    )
+    if enable_deepagent and args.workflow != "main_agent":
+        console.print(
+            "[red]--deepagent adds a worker only to the main_agent workflow. "
```
keceli, line 546: This turns a previously-ignored config setting into a fatal startup error.

With `enable_deepagent = true` in `config.toml`, the user can no longer start the CLI in *any* other workflow. On `main` this is harmless: `interactive_mode` neutralizes the flag via `enable_deepagent and workflow == "main_agent"`, so the session starts and the user can `/workflow main_agent` later to pick it up.

That's not incidental — `test_interactive_deepagent_setting_survives_workflow_switches` covers exactly that path.

Calling `_handle_run` with `deepagent=True, workflow="single_agent", interactive=True` starts the REPL on `main` and exits 2 here.

Since the flag is already neutralized downstream, gating on it only when it was typed on the command line (`cli_deepagent is True`) would keep the new error for the case it's meant for.

src/chemgraph/cli/commands.py
```diff
@@ -1164,6 +1257,7 @@ def interactive_mode(
         if workflow == "main_agent"
         else None
     )
+    standalone_thread_id = _next_thread_id() if workflow == "deep_agent" else None
```
keceli, line 1260: Reusing one thread across REPL queries re-persists the whole accumulated transcript on every turn.

`finalize_completed_run` calls `_save_messages_to_store(last_state, query)`, and `SessionStore.save_messages` is a bare append-only `INSERT` with no dedup. Because the thread is reused, `last_state["messages"]` carries all prior turns, so three queries stored 12 rows for 6 unique messages: `[q1,a1, q1,a1,q2,a2, q1,a1,q2,a2,q3,a3]`. Row count grows O(n²) in turns.

That corrupts `chemgraph sessions show`, `query_count`, and `build_context_summary` — which `--resume` feeds to the next run, so the model gets the same exchange repeated back several times.

Pre-PR this couldn't happen: every standalone workflow got a fresh thread per query. Persisting only the messages added since the last turn would keep the process-local context benefit without the duplication.

src/chemgraph/agent/llm_agent.py
```diff
@@ -961,37 +1081,44 @@ async def _stream_until_interrupt(stream_input, cfg):
...
+                    found_interrupts.append(
+                        PendingInterrupt(
+                            id="",
+                            payload={
+                                "question": "The workflow needs your input."
+                            },
+                        )
+                    )
 
             # Double-check the checkpoint for pending interrupts that
```
keceli, line 1095: Dropping the old `if interrupt_val is None:` guard means the `GraphInterrupt` fallback and the checkpoint scan can both contribute records for the *same* pause.

When `astream` raises a bare `GraphInterrupt()` with no args, the handler appends a synthetic `PendingInterrupt(id="", ...)`. The snapshot scan then adds the real one with its actual id. Their dedup keys differ, so two records survive, and `len(pending) > 1 and any(not pending.id)` raises `RuntimeError: Multiple pending interrupts require stable IDs.`

On `main` the same workflow resumes to completion with a handler configured, or raises the resumable `HumanInputRequired` without one. Here it's fatal.

The old guard suppressed the snapshot scan precisely when the stream had already produced an interrupt. If the scan needs to be unconditional to catch interrupts the stream misses, the synthetic no-id record should be dropped whenever the scan yields anything real.

src/chemgraph/cli/commands.py
```diff
@@ -46,6 +52,7 @@
 ALL_WORKFLOW_TYPES = [
     "single_agent",
     "main_agent",
+    "deep_agent",
```
keceli, line 55: Adding `deep_agent` here makes host-shell access reachable from the plain interactive menu, with no flag and no explicit workspace.

`ALL_WORKFLOW_TYPES` is the `choices=` list for the workflow prompt at commands.py:1197 and for `/workflow`. Run `chemgraph run --interactive` from any directory, pick `deep_agent`, and `uses_deepagent` becomes true with `deepagent_workspace=None` — so `_create_experimental_deepagent_backend(None)` resolves `root = Path.cwd()` and builds a `LocalShellBackend` over whatever directory the user happened to be in.

Before this PR, host-shell access required typing `--deepagent` explicitly.

The asymmetry is what stands out: the headless path guards this exact case ("Headless deep_agent runs require an explicit --deepagent-workspace PATH", main.py:564), and every README/docs example passes `--deepagent-workspace`. The CWD fallback is undocumented, and the only remaining gate is the y/N confirm panel.

Requiring an explicit workspace for the interactive menu path too — or at least naming the resolved root in the confirm prompt — would close the gap.

src/chemgraph/agent/llm_agent.py
```diff
@@ -217,30 +245,57 @@ def __init__(
...
+                "deepagent_skills requires enable_deepagent=True or "
+                "workflow_type='deep_agent'."
+            )
+        if deepagent_auto_approve and workflow_type != "deep_agent":
+            raise ValueError(
+                "deepagent_auto_approve is supported only for the deep_agent "
+                "workflow."
+            )
+        if checkpointer is not None and workflow_type not in {
+            "main_agent",
```
keceli, line 284: Allowing `checkpointer=` for `deep_agent` exposes a sync/async mismatch, because unlike `main_agent` this workflow is driven through the synchronous `run` / `write_state` / `get_state` path.

With an async saver:

```python
async with AsyncSqliteSaver.from_conn_string(db) as s:
    agent = ChemGraph(workflow_type="deep_agent", checkpointer=s,
                      return_option="state", log_dir=d)
    await agent.run(...)
```

the graph completes, then prints `Error with write_state: Synchronous calls to AsyncSqliteSaver are only allowed from a different thread` (log dir ends up empty), then raises `asyncio.InvalidStateError` out of `finalize_completed_run`. With `return_option="last_message"` it returns normally and every state log is silently dropped.

The new unconditional `self.workflow.get_state(cfg)` at llm_agent.py:1103 also no-ops inside its bare `except Exception`, so checkpointed approval interrupts are never recovered for an async saver.

`main_agent` avoids all of this by going exclusively through `MainAgentSession`'s `aget_state`. Either reject async savers here with a clear error, or route `deep_agent` through the async state API.

src/chemgraph/registry/agents.py
```diff
@@ -55,6 +55,13 @@ class AgentSpec:
         default_tool_names=_CORE_TOOLS,
         tags=frozenset({"general", "chemistry"}),
     ),
+    AgentSpec(
+        "deep_agent",
```
keceli, line 59: Registering through the registry gives the subagent a name the supervisor prompt never uses.

`as_subagent` returns `spec.name` (agents.py:344), not the alias that was requested — so `as_subagent("deepagent")` registers the task tool under `deep_agent`. But `DEFAULT_MAIN_AGENT_PROMPT` says `deepagent` throughout ("Use `deepagent` for ...", "delegate those tasks to deepagent") and never mentions `deep_agent`.

Following the new `docs/python_api.md:134-144` example verbatim, the model emits `subagent_type="deepagent"` and gets an unknown-subagent error on every delegation.

The docs say this path "uses this same workflow under the stable subagent name `deepagent`", which only holds for `construct_main_agent_graph` — that hardcodes `name: "deepagent"` separately.

Either have `as_subagent` return the requested alias, or add `deep_agent` to the prompt's vocabulary.

src/chemgraph/agent/interrupts.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def deduplicate_interrupts(
+    interrupts: list[PendingInterrupt],
+) -> tuple[PendingInterrupt, ...]:
+    """Deduplicate streamed and checkpointed copies of pending interrupts."""
+    unique: list[PendingInterrupt] = []
+    seen: set[str] = set()
+    for item in interrupts:
+        key = item.id or repr(item.payload)
```
keceli, line 40: This dedup key treats LangGraph's shared placeholder id as if it were stable.

`Interrupt(value=...)` built without an explicit id gets langgraph 1.2.9's `_DEFAULT_INTERRUPT_ID = "placeholder-id"` (types.py:530). Two distinct interrupts therefore share a key and collapse into one — and because that string is truthy, `any(not pending.id)` returns `False`, so the stable-ID guards in `run` and `run_query` both pass. The second approval request is silently dropped rather than refused, letting one decision stand in for two actions.

The `repr()` fallback for genuinely empty ids has the mirror problem: it's dict-order dependent, so `{'a':1,'b':2}` and `{'b':2,'a':1}` produce different keys (spurious multi-interrupt abort), while two distinct interrupts with equal payloads collapse to one.

Runtime task interrupts use `Interrupt.from_ns` hashes, so the practical trigger is narrow — but treating `placeholder-id` as unset would make the existing guards do their job.

src/chemgraph/cli/commands.py
```diff
@@ -233,18 +256,37 @@ def initialize_agent(
     """
     # Resolve workflow alias before initializing.
     workflow_type = resolve_workflow(workflow_type)
+    deepagent_skills = normalize_skill_sources(deepagent_skills)
```
keceli, line 259: Normalizing before any error handling turns user input errors into raw tracebacks.

A TOML `deepagent_skills = "/workspace/.agents/skills/"` (bare string — truthy, so the guard in main.py passes) reaches here and the run dies after the banner with an unhandled `TypeError: skills must be a sequence of path strings, not a string.` `--deepagent-skill ""` gives an uncaught `ValueError` the same way. `_handle_run` wraps `run_query` but not `initialize_agent`, and every other user-input error in this function exits cleanly via `console.print` + `sys.exit(2)`.

The same ordering bites the Python API: because llm_agent.py normalizes before all four workflow validators, `ChemGraph(workflow_type="single_agent", deepagent_skills="/x/")` raises that `TypeError` instead of the intended `ValueError("deepagent_skills requires enable_deepagent=True...")`, so the actual misuse is masked.

Moving the normalize call after the workflow checks fixes both.

src/chemgraph/cli/commands.py
```diff
@@ -448,22 +494,37 @@ def run_query(
...
+        if len(pending_interrupts) > 1:
+            if any(not pending.id for pending in pending_interrupts):
+                console.print(
+                    "[red]Multiple pending interrupts do not expose stable "
+                    "IDs and cannot be resumed safely.[/red]"
+                )
+                return None
+
+        answers = [
+            _prompt_for_interrupt(pending.payload)
```
keceli, line 514: This comprehension sits outside the `try:` block that starts at line 583, so the `ValueError`s `_prompt_for_interrupt` raises escape `run_query` entirely.

With a tool-review payload whose `allowed_decisions` lacks approve/reject (e.g. a custom `interrupt_on={"execute": {"allowed_decisions": ["edit"]}}`), `_prompt_for_interrupt` raises `ValueError: Deep Agent action 'execute' does not allow approve/reject.` It propagates past `_handle_run` and the CLI dies with a traceback instead of the `[red]Error processing query: ...[/red]` path. Same for the `Invalid Deep Agent approval request.` branch.

docs/python_api.md
```diff
@@ -73,6 +73,79 @@ For CLI use, the equivalent is:
...
+
+```python
+from deepagents.backends import LocalShellBackend
+
+from chemgraph.agent.llm_agent import ChemGraph
+
+agent = ChemGraph(
+    model_name="claude-sonnet-4-20250514",
+    workflow_type="deep_agent",
+    deepagent_backend=LocalShellBackend(
```
keceli, line 88: This example can't run the use case it demonstrates.

`LocalShellBackend` here gets no `env` and no `inherit_env`, so `execute` runs with an empty environment: `pytest --version` comes back `exit_code=127, "python: not found"`.

The CLI's `_create_experimental_deepagent_backend` avoids this by passing an allowlisted env alongside `inherit_env=False`. Worth mirroring that here, or at least pointing readers at it.

Related, since it's the same snippet's assumption: `DEFAULT_DEEPAGENT_PROMPT` tells the model `/workspace` is the project root unconditionally, but `_normalize_backend` only creates that mount for `virtual_mode=True` local backends. With `virtual_mode=False`, a write to `/workspace/x` escapes `root_dir` to the literal host path.

keceli (commented):
Thanks for the fast turnaround — I verified all four items from the first pass on `b4fc462`:

- The fingerprint guard is exactly right, and reconstructing the pre-skills hash in the regression test is a better check than I asked for.
- The `inspect.signature` binding for `human_input_handler` also fixes async *callable objects*, which the old `asyncio.iscoroutinefunction` check missed — a gap I'd noticed but didn't raise. Good catch.
- Public seams and the `--help` wording both read well.
- Agreed on leaving `DEFAULT_DEEPAGENT_PROMPT` alone; the pin plus the asserting test are a reasonable upgrade gate.

CI is green across all 39 checks now, Windows included.

Second pass below. A fuller automated sweep finished after I posted the first review and turned up more than my manual read did; I've verified each of these against `b4fc462` and they're all still live. Sorry for the two-stage delivery — these should have come with the first round.

Nothing here is caught by the current suite (it passes on all three platforms), so the green checks aren't evidence against any of it.

### Roughly in severity order

**Breaks users who never touch Deep Agent** — the config-file handling (`main.py:546`, `main.py:564`) turns two previously-inert settings into hard `exit(2)`s, one of them triggered by the `deepagent_skills` line this PR adds to `config.toml`. `main.py:546` also contradicts `test_interactive_deepagent_setting_survives_workflow_switches`.

**Deep Agent unusable for real tasks** — the interrupt cap (`llm_agent.py:1176`) aborts any multi-file job partway with the workspace already mutated, and the double-mount (`deep_agent.py:64`) shows the model every file twice. The latter also regresses today's `main_agent --deepagent` path.

**Data and state** — REPL thread reuse against append-only `save_messages` (`commands.py:1260`) grows session rows O(n²) and corrupts `--resume`; the unconditional snapshot scan (`llm_agent.py:1095`) can turn a resumable pause into a fatal error.

**Safety** — `deep_agent` is now reachable from the plain interactive menu with no flag, defaulting the host-shell root to CWD (`commands.py:55`).

The rest are smaller: an alias mismatch that breaks the new docs example, two unhandled-traceback paths, a dedup key that can silently drop an approval, and a docs snippet that can't work as written.

Happy to pair on any of these if it's easier than a comment thread — several are one-liners.

---

🤖 Findings generated with [Claude Code](https://claude.com/claude-code) (Opus 5) and posted by @keceli.

src/chemgraph/agent/llm_agent.py
```diff
@@ -802,6 +892,31 @@ def _save_messages_to_store(self, last_state: dict, query: str) -> None:
...
+        self.write_state(config=config)
+
+    def finalize_completed_run(
+        self,
+        last_state: dict,
+        config: dict,
+        query: str,
+    ) -> Any:
+        """Persist a completed run and return the configured result shape."""
+        self._save_messages_to_store(last_state, query)
```
keceli, line 908: Heads-up from #232, which is merging now: this call needs a `thread_id`.

#232 changes `_save_messages_to_store` to append only messages not yet saved for a given thread, tracked per normalized thread ID:

```python
def _save_messages_to_store(self, last_state, query, *, thread_id: str = "1") -> None:
```

`run()` passes the real thread through, but this call site doesn't, so it silently lands in the `"1"` bucket. That matters more here than anywhere else, because this PR introduces `standalone_thread_id` — the reused deep_agent REPL thread — so every thread in that REPL would share one dedup bucket.

Graph-assigned message IDs still discriminate correctly, but #232's fallback key for messages without one is `("position", position, role, content, tool_name)`. Those collide across threads: after `/model` or `/workflow` regenerates `standalone_thread_id`, the new thread restarts at position 0, and a repeated first question would match the old thread's key and be silently dropped. The set also never resets for the life of the REPL.

`config` is already in scope here, so:

```python
self._save_messages_to_store(
    last_state, query, thread_id=str(config["configurable"]["thread_id"])
)
```

Worth checking after you rebase on main, since the parameter won't exist until #232 lands.

---

🤖 Noted with [Claude Code](https://claude.com/claude-code) (Opus 5) and posted by @keceli.

keceli (commented):
One follow-up from #232 (merging now): the new `finalize_completed_run` needs to thread its `thread_id` into `_save_messages_to_store`, or the deep_agent REPL path opts out of that fix. Inline.

src/chemgraph/graphs/deep_agent.py
tdpham2: Confirmed and fixed in 5b6497a81. Host files are now mounted only under `/workspace/`; the default file namespace uses checkpoint state while shell execution delegates to the original local backend. A plain StateBackend default would remove the execute tool, so the adapter retains LocalShellBackend identity and shell-path guidance. Tests cover sync/async glob and grep, root listing, checkpoint-only files, mounted writes, shell execution, and the existing prompt mapping upgrade gate.

src/chemgraph/agent/llm_agent.py
tdpham2: Fixed in 3bbd63945. Structured action approvals no longer consume the ten-question budget in ChemGraph.run, standalone CLI resume, or main-agent CLI operation. Ordinary clarification questions retain the limit. Regressions complete twelve actual file writes through both the Python API and CLI, handle eleven concurrent graph approvals, and verify that excessive questions still stop.

src/chemgraph/cli/main.py
tdpham2: Confirmed and fixed in 97d27591f. Saved Deep Agent settings no longer block unrelated workflows. Interactive sessions retain them for later workflow switches; unrelated headless runs omit them. Explicit incompatible flags still fail. Workflow selection now follows explicit CLI value, then TOML, then single_agent, including the deepagent alias. Parser-to-dispatch tests cover these cases with both enable_deepagent values.

src/chemgraph/cli/main.py
tdpham2: Fixed in 97d27591f. The incompatible-workflow error now applies to an explicitly supplied --deepagent flag. A saved enable_deepagent setting survives interactive startup on another workflow and is activated only when main_agent is selected. The existing workflow-switch regression remains intact, and new tests exercise the real argument parser and config merge.

src/chemgraph/cli/commands.py
tdpham2: Confirmed. Rebasing onto main brings in #232's append-only-unsaved-message fix. I preserved it through the finalization refactor and pass the normalized thread ID on both synchronous and asynchronous completion paths. Existing cumulative-history and failed-write retry tests pass; new regressions verify repeated idless messages are retained across distinct threads without being duplicated within one thread.

src/chemgraph/agent/llm_agent.py
tdpham2: Reproduced and fixed in 3bbd63945. Pending-request collection now prefers checkpoint records over streamed copies, and creates a synthetic fallback only if no real request is available. The standalone runner, CLI resume, and main-agent snapshot handling share the collector. Tests cover a bare GraphInterrupt with a real checkpoint both with and without a handler, and preserve HumanInputRequired when best-effort checkpoint logging fails.

src/chemgraph/agent/llm_agent.py
tdpham2: Reproduced the AsyncSqliteSaver logging failure and InvalidStateError. Fixed in 3bbd63945 by using asynchronous snapshot reads during runs and CLI resumes. The new apersist_run_state and afinalize_completed_run methods share snapshot serialization with the existing synchronous methods. Real AsyncSqliteSaver tests verify interrupted and completed snapshots and both return modes. The docs explain keeping caller-owned savers open on their event loop.

src/chemgraph/registry/agents.py
tdpham2: Confirmed the naming mismatch and fixed it in 5b6497a81. Registry composition keeps the canonical name deep_agent; the supervisor now recognizes that name and instructs task calls to use the exact registered name. The built-in worker remains deepagent. The Python docs distinguish the two, and a real registry-composed supervisor test successfully delegates under the returned canonical name.

src/chemgraph/agent/interrupts.py
tdpham2: Fixed in 3bbd63945. placeholder-id is normalized as missing, and payload repr is no longer a deduplication key. Distinct anonymous requests are retained, even with equal payloads or different dict key order. The collector avoids counting stream/checkpoint copies twice, and concurrent requests without stable IDs are rejected before the handler receives any approval. Tests cover both normalization and the runner's refusal to resume ambiguous requests.

src/chemgraph/cli/commands.py
tdpham2: Fixed in 3bbd63945 and 97d27591f. Workflow compatibility is checked before normalization in the Python API. The CLI catches TypeError/ValueError from skill validation, reports invalid active settings as input errors with exit code 2, and reports interactive initialization errors without enabling the backend. Tests cover a bare string, empty entries, non-string entries, and incompatible Python API usage.

src/chemgraph/cli/commands.py
tdpham2: Fixed in 3bbd63945. Approval collection now runs inside error handling for both standalone CLI resumes and main-agent operations. Tests verify malformed action requests and edit-only decision policies produce a readable error and return without an uncaught traceback.

docs/python_api.md
tdpham2: Confirmed the missing environment and updated the example in 5b6497a81 with an explicit PATH/Python-environment allowlist and inherit_env=False. I left the conditional prompt wording unchanged: it already applies /workspace guidance only when that virtual mount is available. Non-virtual/custom backends retain their existing semantics, with regression coverage for pass-through behavior.

src/chemgraph/agent/llm_agent.py
tdpham2: Addressed during the rebase and retained in both finalization methods. _save_messages_to_store receives str(config["configurable"]["thread_id"]) on normal and resumed completion. New tests finalize identical idless snapshots under 7, "7", and a new thread: the first two share one dedup bucket, while the new thread retains its own exchange. No transcript schema migration is required.

src/chemgraph/cli/commands.py
tdpham2: The resolved root was already displayed in the host-shell warning immediately before the default-no confirmation. I retained that behavior and documented the interactive CWD default in 97d27591f. New tests confirm the exact resolved root is displayed and declining prevents backend creation. Headless runs still require an explicit workspace and the approval-skip flag.

keceli (approved):

PR #222, opened 2026-08-27
fix: start the Streamlit UI on macOS and contain native TBLite crashes

keceli:
## Problem

Two related failures with native calculator runtimes:

1. **The Streamlit UI cannot start on macOS.** Importing `chemgraph.schemas.ase_input` for schema discovery eagerly imported PyTorch (via `fairchem_calc` / `mace_calc`) and TBLite. The macOS TBLite and PyTorch wheels each bundle their own `libomp.dylib`; loading both into one interpreter raises `SIGABRT` before the app renders.

2. **A TBLite crash takes the whole app down.** A native OpenMP loader abort is not a Python exception, so no `try/except` can contain it — the Streamlit process dies instead of reporting a tool error.

## Changes

**Defer native imports** (`5d0625b`)
- `mace_calc` / `fairchem_calc` / `tblite_calc` import torch, mace, fairchem, and tblite inside the methods that use them rather than at module scope.
- `FAIRChemCalc.device` uses `default_factory` so torch is not imported to compute a field default.
- Availability probing uses `find_spec("fairchem")` instead of `find_spec("fairchem.core")`: resolving a submodule spec imports the parent package (and transitively torch), which defeated the point of the gate.
- Missing engines now raise `ImportError` naming the extra to install, replacing a module-scope `logging.warning` that left the calculator name undefined and surfaced later as a confusing `NameError`.

**Contain native crashes** (`3c951f9`)
- TBLite runs in a fresh interpreter (`chemgraph/tools/_ase_worker.py`) via `_run_ase_core_isolated`. A worker killed by a signal becomes a structured `CalculatorProcessError` result, with an added note when the stderr shows an OpenMP runtime conflict.
- `run_ase_core` is now a thin dispatcher; the existing body moved to `_run_ase_core_in_process` unchanged.
- IR runs are rejected up front when the calculator does not implement `dipole`, instead of failing partway through a vibrational analysis.

## Verification

- Full suite on the branch merged into `main`: **690 passed, 29 skipped**. The 10 failures are identical to those on `main` at `d1dcdcb` in the same environment (pre-existing; unrelated to this branch).
- The eager-import fix was checked against tripwire packages that raise on import: `main` fails with `RuntimeError: EAGER IMPORT of torch`, this branch imports cleanly with all seven calculators still discovered.
- The isolation path was exercised end-to-end (subprocess spawn, `PYTHONPATH`, result round-trip, artifact writing into `CHEMGRAPH_LOG_DIR`) for the `opt` and `vib` drivers.
- Confirmed the parent→worker JSON round-trip preserves `TBLiteCalc` rather than coercing it into another member of the non-discriminated calculator union, and that every TBLite alias (`xtb`, `GFN2-xTB`, …) normalizes to `calculator_type="TBLite"` so isolation always triggers.

## Notes for reviewers

Non-blocking observations, listed for the record:

- Isolation costs ~0.4 s of interpreter startup per TBLite call (measured). Negligible interactively; noticeable for large TBLite ensembles or scans, where it can dominate a sub-second single-point.
- `subprocess.run` has no timeout, so a hung worker hangs the caller. Not a regression — an in-process hang did the same — but the subprocess boundary now makes a timeout easy to add.
- `MPLCONFIGDIR` is redirected into a per-call temp dir, so an isolated TBLite **IR** run rebuilds the matplotlib font cache each time.
- If `json.dumps` fails in the worker (e.g. a multi-element NumPy array, which `_json_default` raises `ValueError` on rather than `TypeError`), the write happens outside the worker's `try`, so the failure degrades to an opaque "exited with status 1".

### DIFF
--- a/src/chemgraph/schemas/ase_input.py
+++ b/src/chemgraph/schemas/ase_input.py
@@ -57,7 +57,7 @@ def _command_available(command_name: str, env_var_name: str) -> bool:
     return bool(os.environ.get(env_var_name)) or shutil.which(command_name) is not None
 
 
-if not _engine_available("fairchem.core"):
+if not _engine_available("fairchem"):
     FAIRChemCalc = None
 
 if not _engine_available("mace"):
--- a/src/chemgraph/schemas/calculators/fairchem_calc.py
+++ b/src/chemgraph/schemas/calculators/fairchem_calc.py
@@ -1,15 +1,14 @@
+import logging
+from typing import Any, Dict, Optional
+
 from pydantic import BaseModel, Field, model_validator
 
-from typing import Any, Optional, Dict
-import torch
-import logging
 
-try:
-    from fairchem.core import FAIRChemCalculator, pretrained_mlip
-except ImportError:
-    logging.warning("fairchem is not installed. .")
-    FAIRChemCalculator = None
-    pretrained_mlip = None
+def _default_device() -> str:
+    """Choose a FairChem device without importing PyTorch at module import time."""
+    import torch
+
+    return "cuda" if torch.cuda.is_available() else "cpu"
 
 
 class FAIRChemCalc(BaseModel):
@@ -57,7 +56,7 @@ class FAIRChemCalc(BaseModel):
         default="uma-s-1p1", description="Model names. Options are 'uma-s-1p1' and 'uma-m-1'"
     )
     device: str = Field(
-        default="cuda" if torch.cuda.is_available() else "cpu",
+        default_factory=_default_device,
         description="Computation device to use, either 'cpu' or 'cuda'.",
     )
     inference_settings: str = Field(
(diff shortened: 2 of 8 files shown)


tdpham2:
I think this PR currently combines several independent issues we may need to evaluate separately.

1. Moving the Torch, MACE, FairChem, and TBLite runtime imports closer to calculator construction makes sense.

2. TBLite version and packaging

The crash does not appear to be a general ChemGraph/TBLite problem. I'm aware of this during my earlier attempt to upgrade all ChemGraph's dependencies. It is associated with newer TBLite wheel (0.7.0). Our currently supported and tested TBLite version is 0.4.0, which works in the same process as MACE/UMA without this crash. Versions between these releases have also had known wheel/runtime problems, which is why I still pin 0.4.0. 

Before adding process isolation, we should make our dependency policy consistent:

- The calculators extra in pyproject.toml pins 0.4.0.
- The main Dockerfile pins 0.4.0.
- environment.yml leaves TBLite unpinned.
- Dockerfile.arm installs 0.5.0.

We should align these installation paths. If ChemGraph does not currently support TBLite 0.7.0, we should not add a substantial runtime workaround solely to accommodate it.

3. _run_ase_core_isolated

The implementation isolates every TBLite calculation on every operating system and TBLite version. It adds another worker entry point, JSON serialization, temporary files, PYTHONPATH manipulation, process startup overhead, and additional failure modes.

If we want to support and pin TBLite == 0.7.0, We should spend sometimes making sure this implementation works across platforms (not just MacOS). This also includes running evaluation to make sure there is no numerical difference/issues between 0.4.0 and 0.7.0.

PR #226, opened 2026-09-01
Make MACE-Polar the default calculator

keceli:
## Summary
- make `mace_polar` with `polar-1-m` the default calculator
- add PolarMACE dipole-moment support and related charge, spin, and field metadata
- use one-based vibration-mode labels in dropdowns while hiding mode indices from plots and animations

## Testing
- focused calculator and dipole tests passed
- focused UI tests passed
- Ruff passed for all tracked Python files
- core test suite passed with documented exclusions for tblite and sandbox-sensitive logging tests


src/chemgraph/schemas/calculators/mace_calc.py
```diff
@@ -106,18 +107,26 @@ class MaceCalc(BaseModel):
...
-        Type of calculator to use. Options: 'mace_mp' (default), 'mace_off', or 'mace_anicc'
+        Type of calculator to use. Options: 'mace_polar' (default), 'mace_mp',
+        'mace_off', or 'mace_anicc'.
     model : str or Path, optional
         Name or path to the model file. If None, uses default model for selected calculator type.
     device : str, optional
         Device to use for calculations ('cpu' or 'cuda' or 'xpu'), by default 'cpu'
     default_dtype : str, optional
         Default data type for the model, by default 'float64'. Use 'float32' if device is 'xpu'.
+    charge : int, optional
```
tdpham2, line 124: This is needed, but FYI this will affect the evaluation benchmarks (additional prompts added to the tools). For large model should be fine. For small/weak model with affect their accuracy in general.

The current benchmark questions should be fine (no need to change) since we specified "medium-mpa-0" model in the prompts

environment.yml
```diff
@@ -32,7 +32,8 @@ dependencies:
     - pubchempy==1.0.5
     - numexpr==2.11.0
     - pymatgen==2025.3.10
-    - mace-torch==0.3.13
+    - mace-torch>=0.3.16
+    - graph-longrange @ git+https://github.com/WillBaldwin0/graph_electrostatics.git@v0.4.0
```
tdpham2, line 36: Is this `graph-longrange` used internally in MACE-polar, or where is it used? I don't see an import for this

keceli: Yes, it is required for MACE-POLAR-1. See https://mace-docs.readthedocs.io/en/latest/guide/polar_mace.html#installation.

src/chemgraph/schemas/calculators/mace_calc.py
keceli: That's ok.

PR #228, opened 2026-09-03
Use connectivity_smiles for PubChem name lookups

### LINKED ISSUE #200: [DEV] Update cheminformatics tools
When running molecule_name_to_smile, I'm seeing this deprecation warning:

```
/Users/tpham2/work/projects/ChemGraph/src/chemgraph/tools/cheminformatics_core.py:97: 
PubChemPyDeprecationWarning: canonical_smiles is deprecated: Use connectivity_smiles instead
  smiles = comps[0].canonical_smiles
```

We should update .canonical_smiles to .connectivity_smiles.

reowszer:
## Summary

`molecule_name_to_smiles` emitted a `PubChemPyDeprecationWarning` on every call: pubchempy 1.0.5 deprecates `canonical_smiles` in favour of `connectivity_smiles`. The first commit does that rename, with no behaviour change: the deprecated property returns `self.connectivity_smiles`.

The second commit is an optional follow-up. Connectivity SMILES carries no stereochemistry, so `"L-alanine"` and `"D-alanine"` both resolve to `CC(C(=O)O)N`: the requested enantiomer is already lost in the returned string. Embedding that string still produces an assigned centre, S or R depending on the seed, and at `smiles_to_3d`'s default seed both names give S: a request for D-alanine produces the L configuration. An opt-in `include_stereochemistry` flag returns PubChem's isomeric SMILES instead. It defaults to `False`, so existing callers are unaffected.

Happy to drop the second commit if you would rather keep this to the deprecation fix.

## Related issues

Closes #200

## Type of change

- [x] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Chore / refactor / CI

## How was this tested?

`pytest tests/test_tools.py`: 12 passed, run with `-W error::pubchempy.PubChemPyDeprecationWarning` to confirm the warning is gone. New test asserts both the default and the flag.

Against live PubChem:

```
L-alanine    off='CC(C(=O)O)N'   on='C[C@@H](C(=O)O)N'
D-alanine    off='CC(C(=O)O)N'   on='C[C@H](C(=O)O)N'
```

Embedded to 3D through `smiles_to_3d`, the default path gives the S centre for both names, while the flag separates them into S and R.

`pytest tests/ -k "not tblite"`: 21 failed, 731 passed, 30 skipped. A clean `main` (`ceee78e`) fails the same 21 tests, name for name. This branch adds no new failures.

## Checklist

- [x] `ruff check .` passes

`pytest` unticked: 21 pre-existing failures on `main`, unchanged by this PR. Docs unticked: no page under `docs/` documents this tool's arguments, so the new flag is described in its docstrings and the MCP tool description only.

### DIFF
--- a/.opencode/skills/chemgraph/SKILL.md
+++ b/.opencode/skills/chemgraph/SKILL.md
@@ -130,7 +130,7 @@ def molecule_name_to_smiles(name: str) -> str:
     comps = pcp.get_compounds(name.strip(), "name")
     if not comps:
         raise ValueError(f"No PubChem compound found for: {name}")
-    return comps[0].connectivity_smiles
+    return comps[0].connectivity_smiles  # .smiles keeps stereochemistry
 ```
 
 ## How to add a new MCP server tool
--- a/.opencode/skills/chemgraph/SKILL.md
+++ b/.opencode/skills/chemgraph/SKILL.md
@@ -130,7 +130,7 @@ def molecule_name_to_smiles(name: str) -> str:
     comps = pcp.get_compounds(name.strip(), "name")
     if not comps:
         raise ValueError(f"No PubChem compound found for: {name}")
-    return comps[0].canonical_smiles
+    return comps[0].connectivity_smiles
 ```
 
 ## How to add a new MCP server tool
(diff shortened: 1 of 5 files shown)


tdpham2:
LGTM!

PR #230, opened 2026-09-03
Add committee voting and calibrated confidence to OCSR

reowszer:
## Summary

This PR adds committee voting to `image_to_smiles`, for a committee of any size, and a measured probability with a confidence interval behind each result. The interval comes from a calibration table, and a CLI refits that table from the user's own labelled images, so the statistic and its interval describe the committee and the images actually in use.

`ensemble=True` reads the image with every specialist that can run and votes. How much the models agree is measurable, so the vote pattern looks up a probability measured on labelled images. Nothing fixes the committee at four: the vote reports a pattern for whatever set of models ran, and the validator checks a table against the committee it was fitted on. Four models give five patterns:

| agreement | meaning |
|-----------|---------|
| `4` | all four agree |
| `3/1` | three against one |
| `2/1/1` | two agree, two differ |
| `2/2` | an even split |
| `1/1/1/1` | all different |

```python
image_to_smiles_core("molecule.png", ensemble=True)
# {'smiles': 'CC(=O)Oc1ccccc1C(=O)O', 'confidence': ...,
#  'confidence_label': 'unanimous', 'agreement': '4', 'votes': {...}}
```

One table ships as the default, fitted for these four models on RDKit-rendered diagrams. A different committee, or scans and journal crops instead of clean diagrams, can behave differently, so `python -m chemgraph.tools.ocsr_calibrate` refits a table from your own labelled images and `--validate` checks the shipped one against a sample of them.

This builds on #198, which added the specialists and the model selection. `ensemble` defaults to off, so a single-model call runs the same model and returns the same SMILES it did there. The result gains confidence fields, which stay `None` on a single-model call because one model produces no agreement to score.

## Related issues

Follow-up to #198. No open issue.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Chore / refactor / CI

## How was this tested?

`pytest tests/`: 949 passed, 32 skipped, 21 failed. `ruff check .` clean. This branch adds 150 test functions, 218 cases once parametrized. All are hermetic: the backends are stubbed, so no model is loaded and no image is read.

The 21 failures are already on `main` and are unrelated to this PR.

Every figure quoted in the code and the docs was checked against the raw benchmark predictions, and the shipped table regenerates from them.

## Checklist

- [x] `ruff check .` passes
- [x] `pytest tests/ -k "not tblite"`: 949 passed, 30 skipped, 2 deselected, 21 failed. The 21 are already on `main`.

## What changed

- **`ocsr_core.py`** votes the predictions, and loads a calibration table through a validator that refuses one which cannot mean what a caller will assume.
- **`ocsr_tools.py`** runs the specialists, votes the ones that returned a reading, and attaches the confidence. `models_wanted` votes a subset, for when a table covers fewer models than are installed.
- **`ocsr_calibrate.py`** is new: it fits a table from labelled images, and `--validate` scores the shipped one against a sample.

### DIFF
--- a/examples/ocsr/README.md
+++ b/examples/ocsr/README.md
@@ -53,8 +53,9 @@ it. The published accuracies were measured under the other prompt.
 Exact match is over a 722-image benchmark. It ranks the four against each other;
 it does not predict how any of them will do on your images. DECIMER is the default because it is the most accurate of the four.
 
-`list_ocsr_models` reports the same table plus what is actually installed on the
-machine you are on.
+`list_ocsr_models` reports what is actually installed on the machine you are on.
+The accuracies it prints come from the calibration table, so they follow a refit
+(see below); the numbers above are the shipped table's, measured on the benchmark.
 
 ## Installing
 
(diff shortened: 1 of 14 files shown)


tdpham2:
Thanks @reowszer. I'd like to see the improvement of using the committee versus just DECIMER on a larger test set like we discussed before reviewing this.

PR #233, opened 2026-09-05
fix: correct thermochemistry and preserve ASE mode diagnostics

### LINKED ISSUE #141: `thermo` driver crashes for every odd-electron (radical) species: `get_symmetry_number` builds a pymatgen Molecule with spin_multiplicity=1
## Summary

Any thermochemistry calculation on an odd-electron species (OH, CH3, NO, NO2, HO2, ...) raises
`ValueError: Charge of 0 and spin multiplicity of 1 is not possible for this molecule`.

Note that NO2 ("Nitrogen Dioxide") is part of the molecule list in
`scripts/evaluations/input_data.json`, so the bundled evaluation set itself contains
species that cannot pass through the `thermo` driver.

## Root cause

`get_symmetry_number` (`src/chemgraph/tools/ase_core.py`, lines 122-146) converts the ASE
`Atoms` to a pymatgen `Molecule` via `AseAtomsAdaptor.get_molecule(atoms)`. The `Atoms`
object is built without magnetic moments, so the adapter derives `spin_multiplicity = 1`.
For an odd-electron system, charge 0 with multiplicity 1 is impossible, and
`Molecule.set_charge_and_spin` raises.

The crash happens inside the `thermo` branch of the driver (`ase_core.py`, line 569,
`symmetrynumber = get_symmetry_number(final_structure)`), i.e. after the (potentially
expensive) optimization and vibrational analysis have already run.

## Minimal reproduction

```python
from chemgraph.schemas.atomsdata import AtomsData
from chemgraph.tools.ase_core import get_symmetry_number

# hydroxyl radical: 9 electrons
oh = AtomsData(numbers=[8, 1], positions=[[0.0, 0.0, 0.0], [0.0, 0.0, 0.97]])
get_symmetry_number(oh)
```

```
File ".../chemgraph/tools/ase_core.py", line 144, in get_symmetry_number
    molecule = aaa.get_molecule(atoms)
File ".../pymatgen/io/ase.py", line 452, in get_molecule
    molecule.set_charge_and_spin(charge, spin_multiplicity=spin_mult)
File ".../pymatgen/core/structure.py", line 5312, in set_charge_and_spin
    raise ValueError(
ValueError: Charge of 0 and spin multiplicity of 1 is not possible for this molecule
```

Environment: chemgraph 0.5.0, pymatgen 2026.5.18, Python 3.12.

## Suggested fix (verified)

The rotational symmetry number is a purely geometric property; charge/spin bookkeeping is
irrelevant here. Disabling the consistency check in the adapter call is enough:

```python
molecule = aaa.get_molecule(atoms, charge_spin_check=False)
```

Verified locally: with this one-line change, `get_symmetry_number` returns 1 for the OH
radical, and closed-shell results are unchanged. Happy to open a PR if this direction
looks right to you.

(Found while running guarded end-to-end benchmarks on top of ChemGraph — thanks for the
framework and the public evaluation data.)

tdpham2:
## Summary

Monatomic thermochemistry previously returned zero entropy and set enthalpy/Gibbs energy equal to potential energy at every temperature and pressure. Compute the atomic contributions with ASE IdealGasThermo and skip atomic displacement calculations. Geometric symmetry detection permits radicals without changing the calculator's electronic state.

Molecular thermochemistry now uses ASE 3.29's explicit `vib_selection="highest", ignore_imag_modes=False` policy on the complete complex spectrum. Reported frequencies, CSV entries, and mode trajectories follow ASE's retained energies, including correct indices for degenerate modes. Preserve the complete input spectrum, selection/cleanup metadata, and warnings in JSON and HTML. If thermochemistry fails or returns non-finite values, save the completed structure, energy, convergence state, and spectrum and return the result-file path. Failure messages distinguish the saved JSON from empty selected-mode exports and explain that single atoms have no vibrational artifacts.

Require ASE >= 3.29.0 and use 3.29.0 in the maintained environment and both Docker variants. Normalize omitted/null ASE input temperatures to 298.15 K, clean stale atomic artifacts, and validate/display entropy units independently of energy units.

Reject nonpositive TBLite/ORCA spin multiplicities during input validation while preserving existing defaults. Keep the report JavaScript conversion tests and install QuickJS from the optional `test` extra in Ubuntu CI jobs, with an explicit import check.

HTML reports tolerate missing thermochemistry diagnostics and unavailable mode mappings while keeping the structure, energy, and spectra visible. Show unknown mode mappings explicitly, describe the recorded imaginary-mode policy accurately, and keep unsupported entropy units as errors. Handled report failures return `Error:` messages so the agent retries instead of treating a missing report as success.

Single-agent and executor prompts use defaults declared by the selected tool schema, preserve user-supplied values, and avoid inventing defaults. The planner delegates schema validation and default handling to executors. This guidance remains available when human supervision is disabled.

## Scientific behavior and compatibility

- ASE selects the expected number of modes by signed squared energy, then rejects any imaginary modes remaining after selection. Zero-energy modes are retained; non-finite thermochemistry values cause a failure. Selection may exclude imaginary modes before validation, so successful thermochemistry does not establish structural stability. The complete spectrum remains available in JSON and HTML.
- `n_imag` is zero on successful current calculations. Legacy results using `ignore_imag_modes=True` may record modes removed after selection, including zero-energy modes. `raw_imaginary_mode_count` separately records imaginary input modes and does not itself trigger a warning. Preserve ASE warnings and warn when no vibrational modes contribute.
- On molecular thermochemistry failure, the complete input spectrum remains in JSON under `vibrational_frequencies.all_modes`; selected frequencies and mode indices are empty, the selected-mode CSV has no rows, and no selected-mode trajectories are exported.
- Temperature normalization applies to ASE single and ensemble inputs. Validation changes to the legacy `run_mace` wrappers are deferred to a separate PR; MACE calculations through the ASE interface use the ASE input schema.
- Existing energy/entropy fields and units are preserved. New mode indices are zero-based; HTML displays original mode numbers starting at 1. Legacy reports remain supported.
- Standalone vibration and IR selection is unchanged. Missing calculator multiplicity still warns and assumes a singlet; automatic ground-state spin inference remains outside this change.
- Ideal-gas thermochemistry expects isolated, nonperiodic molecules with unwrapped coordinates. Periodic inputs are rejected; periodic molecule reconstruction is outside this change.

## Related issues

Closes #141. Release prerequisite for #227.

## Type of change

- [x] Bug fix
- [x] Docs
- [x] Dependency / Docker update

## How was this tested?

- `ruff check .` and `git diff --check` passed.
- Core test gate at `caf189473`: `pytest tests/ -k "not tblite"` — **995 passed, 21 skipped, 2 deselected**, run locally with ASE 3.29.0, QuickJS 1.19.4, and Academy/Parsl/Globus extras.
- Thermochemistry/report regressions pass in the full suite above, including all three QuickJS JavaScript conversion cases. Ubuntu `test-extras` jobs for Python 3.11/3.12 now install the `test` extra and verify the QuickJS import before pytest; the results above are local verification, not a claim of completed CI.
- Recomputed thermochemistry from the saved methane 500 K spectrum: `n_imag=0`, `raw_imaginary_mode_count=3`, `warnings=[]`; retained mode indices and H/S/G match the saved result (absolute tolerance 1e-12).
- Controlled complex spectra cover recorded Cu4/EMT energies, duplicate energies at the selection boundary, and nontrivial mode ordering. Verify H/S/G and retained spectra against ASE, and verify exact CSV/trajectory indices. Selected imaginary modes (including all-imaginary inputs) and zero energies that produce non-finite values exercise the failure path.
- Constructor, entropy, and non-finite-result failures preserve completed output and a JSON-serializable failure response. Failure-message regressions verify that molecular spectra survive in JSON with empty selected-mode CSVs and no selected-mode trajectories, while atomic failures preserve structure and energy without claiming a vibrational spectrum.
- Thirteen schema-only regression cases cover rejection of zero/negative multiplicities, preservation of valid values, and unchanged TBLite/ORCA defaults.
- Report tests cover missing/null diagnostic metadata, absent or inconsistent mode mappings, full spectra, warnings, partial failures, current and legacy imaginary-mode policies, HTML escaping, unsupported entropy units, preservation of existing reports after validation errors, and actual conversion JavaScript executed locally with QuickJS.
- Actual successful and failed report-tool outputs exercise both routing functions, verifying completion after success and retry after failure.
- Prompt regression tests verify that schema-default guidance and planner delegation survive removal of human-routing instructions, including the legacy planner-prompt alias.
- Molecular rerun tests verify preservation of prior artifacts during displacement failures, prior trajectories during mode-write failures, and prior IR artifacts during IR-calculation failures.
- Single-atom IR skips dipole validation because no vibrational modes are calculated; molecular IR still requires a dipole-capable calculator. Both behaviors are covered by the passing suite.
- Both Dockerfiles passed shell and embedded-Python syntax checks after rebasing onto main. The final validation block pins ASE 3.29.0 and preserves main's NWChem/TBLite/graph-longrange checks and `python -m pip check`. **Docker image builds were not run because the local Docker daemon is unavailable.**

## Checklist

- [x] Targets main; no direct or force push to main
- [x] Ruff and the core test gate pass
- [x] Added hermetic scientific and result-preservation regressions
- [x] Updated schema descriptions, calculator documentation, and Docker documentation

### DIFF
--- a/docs/calculators.md
+++ b/docs/calculators.md
@@ -36,6 +36,13 @@ Calculator-backed tools cover operations such as:
 Support depends on the selected calculator. A valid property for one engine may
 not exist for another.
 
+Ideal-gas thermochemistry uses the requested temperature and pressure. Enthalpy
+and Gibbs energy are reported in eV; entropy is reported in eV/K with a separate
+`entropy_unit` field. HTML reports also support kJ/(mol K) and kcal/(mol K) for
+entropy. Single atoms include translational and electronic-spin contributions
+without running finite-difference vibrations. Calculator multiplicity determines
+the electronic-spin contribution.
+
 ## EMT for setup checks
 
 EMT is lightweight and requires no download, making it a useful plumbing test.
--- a/docs/calculators.md
+++ b/docs/calculators.md
@@ -46,10 +46,12 @@ results without this field, and reject other declared entropy units. The shared
 to kJ/(mol K) or kcal/(mol K) for the corresponding molar energy selection.
 
 ChemGraph requires ASE >= 3.29.0. For `thermo`, the complete complex spectrum is
-passed to `IdealGasThermo(vib_selection="highest", ignore_imag_modes=True)`.
-ASE selects the expected number of modes by squared energy, then removes any
-remaining imaginary or zero-energy modes. ChemGraph does not apply an additional
-frequency cutoff or convert imaginary frequencies to real ones.
+passed to `IdealGasThermo(vib_selection="highest", ignore_imag_modes=False)`.
+These are ASE's defaults: select the expected number of modes by signed squared
+energy, then reject any remaining imaginary modes. Zero-energy modes are not
+automatically removed; if they yield non-finite thermochemistry, the calculation
+returns a failure. ChemGraph does not apply an additional check of the complete
+spectrum, impose a frequency cutoff, or convert imaginary frequencies to real ones.
 
 Reported thermochemistry frequencies, CSV entries, and trajectories match the
 energies ASE actually used. `vibrational_frequencies.mode_indices` contains their
@@ -60,9 +62,11 @@ spectrum with used/excluded labels. Standalone `vib` and `ir` output is unchange
 
 Thermochemistry metadata records `ase_version`, `vib_selection`,
 `ignore_imag_modes`, `n_imag`, `raw_imaginary_mode_count`, and `warnings`.
-`n_imag` is ASE's cleanup count **after selection**, including zero-energy modes;
-it is not the number of imaginary modes in the complete input. Selection may
-already have excluded imaginary modes even when `n_imag` is zero. Successful
+`n_imag` is zero for successful calculations under this policy. Legacy results
+with `ignore_imag_modes=True` may record modes removed **after selection**,
+including zero-energy modes; HTML reports continue to support that metadata.
+`n_imag` is not the number of imaginary modes in the complete input. Selection
+may already have excluded imaginary modes even when `n_imag` is zero. Successful
 thermochemistry with excluded modes does not establish structural stability.
 The raw imaginary-mode count is diagnostic and does not itself trigger a
 warning. Warnings contain messages emitted by ASE and identify calculations
--- a/docs/calculators.md
+++ b/docs/calculators.md
@@ -45,6 +45,31 @@ results without this field, and reject other declared entropy units. The shared
 **Units** selector converts energy and entropy together; entropy labels change
 to kJ/(mol K) or kcal/(mol K) for the corresponding molar energy selection.
 
+ChemGraph requires ASE >= 3.29.0. For `thermo`, the complete complex spectrum is
+passed to `IdealGasThermo(vib_selection="highest", ignore_imag_modes=True)`.
+ASE selects the expected number of modes by squared energy, then removes any
+remaining imaginary or zero-energy modes. ChemGraph does not apply an additional
+frequency cutoff or convert imaginary frequencies to real ones.
+
+Reported thermochemistry frequencies, CSV entries, and trajectories match the
+energies ASE actually used. `vibrational_frequencies.mode_indices` contains their
+original zero-based ASE indices; `all_modes` preserves every input mode as
+`mode_index`, `energy` (meV), and `frequency` (cm-1), with an `i` suffix for
+imaginary values. HTML displays mode numbers starting at 1 and includes the full
+spectrum with used/excluded labels. Standalone `vib` and `ir` output is unchanged.
+
+Thermochemistry metadata records `ase_version`, `vib_selection`,
+`ignore_imag_modes`, `n_imag`, `raw_imaginary_mode_count`, and `warnings`.
+`n_imag` is ASE's cleanup count **after selection**, including zero-energy modes;
+it is not the number of imaginary modes in the complete input. Selection may
+already have excluded imaginary modes even when `n_imag` is zero. Successful
+thermochemistry with excluded modes does not establish structural stability.
+Warnings identify imaginary input modes and calculations with no vibrational
+contribution. If ASE raises or returns non-finite thermodynamic values, ChemGraph
+returns a failure with `results_file` pointing to the completed structure,
+potential energy, convergence state, and full spectrum; the JSON records
+`success=false` and the error, with no thermochemistry values.
+
 Single atoms skip finite-difference vibrations for `thermo`, `vib`, and `ir`.
 Atomic thermochemistry includes translation and uses the calculator's reported
 multiplicity for the electronic-spin contribution. If no multiplicity is
--- a/docs/calculators.md
+++ b/docs/calculators.md
@@ -36,12 +36,23 @@ Calculator-backed tools cover operations such as:
 Support depends on the selected calculator. A valid property for one engine may
 not exist for another.
 
-Ideal-gas thermochemistry uses the requested temperature and pressure. Enthalpy
-and Gibbs energy are reported in eV; entropy is reported in eV/K with a separate
-`entropy_unit` field. HTML reports also support kJ/(mol K) and kcal/(mol K) for
-entropy. Single atoms include translational and electronic-spin contributions
-without running finite-difference vibrations. Calculator multiplicity determines
-the electronic-spin contribution.
+Ideal-gas thermochemistry uses the requested temperature and pressure (defaults:
+298.15 K and 101325 Pa). ASE single and ensemble inputs also interpret a null
+temperature as 298.15 K; supplied temperatures must be finite and positive.
+Enthalpy and Gibbs energy are reported in eV; entropy is reported in eV/K with a
+separate `entropy_unit` field. HTML reports accept eV/K entropy, including legacy
+results without this field, and reject other declared entropy units. The shared
+**Units** selector converts energy and entropy together; entropy labels change
+to kJ/(mol K) or kcal/(mol K) for the corresponding molar energy selection.
+
+Single atoms skip finite-difference vibrations for `thermo`, `vib`, and `ir`.
+Atomic thermochemistry includes translation and uses the calculator's reported
+multiplicity for the electronic-spin contribution. If no multiplicity is
+reported, ChemGraph logs a warning and assumes a singlet, omitting the
+electronic-spin entropy of open-shell species. This does not infer ground-state
+multiplicities or add spin dependence to a calculator's potential energy.
+Rotational symmetry analysis expects an isolated, unwrapped molecule; periodic
+images are not reconstructed.
 
 ## EMT for setup checks
 
--- a/docs/calculators.md
+++ b/docs/calculators.md
@@ -64,9 +64,11 @@ Thermochemistry metadata records `ase_version`, `vib_selection`,
 it is not the number of imaginary modes in the complete input. Selection may
 already have excluded imaginary modes even when `n_imag` is zero. Successful
 thermochemistry with excluded modes does not establish structural stability.
-Warnings identify imaginary input modes and calculations with no vibrational
-contribution. If ASE raises or returns non-finite thermodynamic values, ChemGraph
-returns a failure with `results_file` pointing to the completed structure,
+The raw imaginary-mode count is diagnostic and does not itself trigger a
+warning. Warnings contain messages emitted by ASE and identify calculations
+with no vibrational contribution. If ASE raises or returns non-finite
+thermodynamic values, ChemGraph returns a failure with `results_file` pointing
+to the completed structure,
 potential energy, convergence state, and full spectrum; the JSON records
 `success=false` and the error, with no thermochemistry values.
 
--- a/src/chemgraph/tools/ase_core.py
+++ b/src/chemgraph/tools/ase_core.py
@@ -248,7 +248,10 @@ def _vibrational_mode_indices(atomsdata: AtomsData, total_modes: int) -> list[in
 
 
 def get_symmetry_number(atomsdata: AtomsData) -> int:
-    """Return the rotational symmetry number using Pymatgen.
+    """Return the rotational symmetry number of an isolated molecule.
+
+    Coordinates must describe an unwrapped molecule; periodic images are
+    not reconstructed for point-group analysis.
 
     Parameters
     ----------
@@ -258,6 +261,9 @@ def get_symmetry_number(atomsdata: AtomsData) -> int:
     -------
     int
     """
+    if len(atomsdata.numbers) == 1:
+        return 1
+
     from pymatgen.symmetry.analyzer import PointGroupAnalyzer
     from ase import Atoms
     from pymatgen.io.ase import AseAtomsAdaptor
@@ -749,6 +755,7 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
         thermo_data: dict = {}
         vib_data: dict = {}
         ir_data: dict = {}
+        ir_plot_path: Optional[str] = None
 
         # --------------------------------------------------------------
         # Vibrational / thermo / IR analysis
@@ -761,17 +768,37 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
                 "frequencies": [],
                 "frequency_unit": "cm-1",
             }
-        if driver in {"vib", "thermo", "ir"} and not (
-            driver == "thermo" and len(atoms) == 1
-        ):
-            logger.info("Starting vibrational analysis (driver=%s)", driver)
-            from ase.vibrations import Vibrations
-            from ase import units
-
-            ir_plot_path: Optional[str] = None
             mol_stem = (
                 Path(input_structure_file).stem if input_structure_file else "mol"
             )
+            # Remove previous modes even when this run has no vibrations.
+            freq_file = Path(_resolve_path(f"frequencies_{mol_stem}.csv"))
+            freq_file.unlink(missing_ok=True)
+            traj_dest_dir = _resolve_path("")
+            stale_traj_pattern = (
+                glob.escape(_resolve_path(f"{mol_stem}_vib.")) + "*.traj"
+            )
+            for stale_traj_file in glob.glob(stale_traj_pattern):
+                os.unlink(stale_traj_file)
+
+            if driver == "ir":
+                ir_data = {
+                    "spectrum_frequencies": [],
+                    "spectrum_frequencies_units": "cm-1",
+                    "spectrum_intensities": [],
+                    "spectrum_intensities_units": "D/Å^2 amu^-1",
+                }
+                for name in (
+                    f"ir_spectrum_{mol_stem}.png",
+                    f"ir_spectrum_{mol_stem}.csv",
+                    f"ir_peaks_{mol_stem}.csv",
+                ):
+                    Path(_resolve_path(name)).unlink(missing_ok=True)
+
+        if driver in {"vib", "thermo", "ir"} and len(atoms) > 1:
+            logger.info("Starting vibrational analysis (driver=%s)", driver)
+            from ase.vibrations import Vibrations
+            from ase import units
 
             with tempfile.TemporaryDirectory(
                 prefix=f"chemgraph_vib_{mol_stem}_"
@@ -798,10 +825,6 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
                     vib_data["frequencies"].append(f"{freq_cm1}{suffix}")
 
                 # Write frequencies CSV
-                freq_file_path = _resolve_path(f"frequencies_{mol_stem}.csv")
-                freq_file = Path(freq_file_path)
-                if freq_file.exists():
-                    freq_file.unlink()
                 with freq_file.open("w", encoding="utf-8") as f:
                     for mode_index, freq in zip(
                         mode_indices, vib_data["frequencies"]
@@ -814,14 +837,8 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
                         n=mode_index, kT=units.kB * 300, nimages=30
                     )
 
-                traj_dest_dir = _resolve_path("")
                 if traj_dest_dir:
                     os.makedirs(traj_dest_dir, exist_ok=True)
-                stale_traj_pattern = os.path.join(
-                    traj_dest_dir, f"{mol_stem}_vib.*.traj"
-                )
-                for stale_traj_file in glob.glob(stale_traj_pattern):
-                    os.unlink(stale_traj_file)
                 for mode_index in mode_indices:
                     traj_file = os.path.join(tmpdir, f"vib.{mode_index}.traj")
                     dest_name = f"{mol_stem}_{Path(traj_file).name}"
@@ -840,11 +857,6 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
 
                     matplotlib.use("Agg")
                     import matplotlib.pyplot as plt
-
-                    ir_data["spectrum_frequencies"] = []
-                    ir_data["spectrum_frequencies_units"] = "cm-1"
-                    ir_data["spectrum_intensities"] = []
-                    ir_data["spectrum_intensities_units"] = "D/Å^2 amu^-1"
 
                     ir_name = os.path.join(tmpdir, "ir")
                     ir = Infrared(atoms, name=ir_name)
@@ -916,7 +928,15 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
                 symmetrynumber = get_symmetry_number(final_structure)
 
             # IdealGasThermo expects total spin S; calculators expose 2S+1.
-            multiplicity = getattr(calc_model, "get_multiplicity", lambda: None)() or 1
+            multiplicity = getattr(calc_model, "get_multiplicity", lambda: None)()
+            if multiplicity is None:
+                logger.warning(
+                    "%s does not report spin multiplicity; assuming a singlet "
+                    "(multiplicity=1) for thermochemistry. Electronic-spin entropy "
+                    "is omitted for open-shell species.",
+                    type(calc_model).__name__,
+                )
+                multiplicity = 1
             thermo = IdealGasThermo(
                 vib_energies=all_energies,
                 potentialenergy=potential_energy,
@@ -925,14 +945,12 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
                 symmetrynumber=symmetrynumber,
                 spin=(multiplicity - 1) / 2.0,
             )
+            enthalpy = float(thermo.get_enthalpy(temperature, verbose=False))
+            entropy = float(thermo.get_entropy(temperature, pressure, verbose=False))
             thermo_data = {
-                "enthalpy": float(thermo.get_enthalpy(temperature=temperature)),
-                "entropy": float(
-                    thermo.get_entropy(temperature=temperature, pressure=pressure)
-                ),
-                "gibbs_free_energy": float(
-                    thermo.get_gibbs_energy(temperature=temperature, pressure=pressure)
-                ),
+                "enthalpy": enthalpy,
+                "entropy": entropy,
+                "gibbs_free_energy": enthalpy - temperature * entropy,
                 "unit": "eV",
                 "entropy_unit": "eV/K",
             }
@@ -1006,6 +1024,12 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
                 ),
             }
         elif driver == "ir":
+            artifact_message = (
+                f"IR plot saved to {os.path.abspath(ir_plot_path)}. "
+                "Normal modes saved as individual .traj files"
+                if ir_plot_path
+                else "Single atoms have no vibrational modes or IR spectrum."
+            )
             return {
                 "status": "success",
                 **energy_metadata,
@@ -1013,8 +1037,7 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
                 "message": (
                     "Infrared computed and returned. "
                     f"Full results (structure, vibrations, thermochemistry and metadata) saved to {abs_output}. "
-                    f"IR plot saved to {os.path.abspath(ir_plot_path) if ir_plot_path else 'N/A'}. "
-                    "Normal modes saved as individual .traj files"
+                    f"{artifact_message}"
                 ),
             }
 
--- a/src/chemgraph/tools/ase_core.py
+++ b/src/chemgraph/tools/ase_core.py
@@ -15,6 +15,7 @@ import os
 import shutil
 import tempfile
 import time
+import warnings
 from pathlib import Path
 from typing import List, Optional
 
@@ -245,6 +246,110 @@ def _vibrational_mode_indices(atomsdata: AtomsData, total_modes: int) -> list[in
 
     num_nonvibrational = 5 if is_linear_molecule(atomsdata) else 6
     return list(range(num_nonvibrational, total_modes))
+
+
+def _vibrational_mode_record(mode_index: int, energy: complex) -> dict:
+    """Format an original ASE mode without hiding small imaginary energies."""
+    from ase import units
+
+    value = energy.imag if energy.imag != 0 else energy.real
+    suffix = "i" if energy.imag != 0 else ""
+    return {
+        "mode_index": mode_index,
+        "energy": f"{1e3 * value}{suffix}",
+        "frequency": f"{value / units.invcm}{suffix}",
+    }
+
+
+def _calculate_thermochemistry(
+    atoms,
+    final_structure,
+    all_energies,
+    potential_energy,
+    calc_model,
+    temperature,
+    pressure,
+) -> tuple[dict, list[int]]:
+    """Let ASE select/clean modes and map its retained energies to ASE indices."""
+    import ase
+    from ase.thermochemistry import IdealGasThermo
+
+    expected_modes = 0 if len(atoms) == 1 else 3 * len(atoms)
+    if len(all_energies) != expected_modes:
+        raise ValueError(
+            f"Expected {expected_modes} input modes, got {len(all_energies)}."
+        )
+    if len(atoms) == 1:
+        geometry, symmetrynumber = "monatomic", 1
+    else:
+        geometry = "linear" if is_linear_molecule(final_structure) else "nonlinear"
+        symmetrynumber = get_symmetry_number(final_structure)
+
+    # IdealGasThermo expects total spin S; calculators expose 2S+1.
+    multiplicity = getattr(calc_model, "get_multiplicity", lambda: None)()
+    if multiplicity is None:
+        logger.warning(
+            "%s does not report spin multiplicity; assuming a singlet "
+            "(multiplicity=1) for thermochemistry. Electronic-spin entropy "
+            "is omitted for open-shell species.",
+            type(calc_model).__name__,
+        )
+        multiplicity = 1
+
+    with warnings.catch_warnings(record=True) as caught:
+        warnings.simplefilter("always", UserWarning)
+        thermo = IdealGasThermo(
+            vib_energies=all_energies,
+            potentialenergy=potential_energy,
+            atoms=atoms,
+            geometry=geometry,
+            symmetrynumber=symmetrynumber,
+            spin=(multiplicity - 1) / 2.0,
+            vib_selection="highest",
+            ignore_imag_modes=True,
+        )
+        retained_energies = thermo.vib_energies
+        enthalpy = float(thermo.get_enthalpy(temperature, verbose=False))
+        entropy = float(thermo.get_entropy(temperature, pressure, verbose=False))
+    gibbs_free_energy = enthalpy - temperature * entropy
+    if not np.all(np.isfinite([enthalpy, entropy, gibbs_free_energy])):
+        raise ValueError("ASE returned non-finite thermochemistry values.")
+
+    # ASE sorts stably and keeps the last modes. Match equal energies from
+    # the end so degenerate modes at the selection boundary keep their IDs.
+    indices_by_energy: dict[float, list[int]] = {}
+    for index, energy in enumerate(all_energies):
+        indices_by_energy.setdefault(float(np.real(energy)), []).append(index)
+    mode_indices = [
+        indices_by_energy[float(energy)].pop() for energy in reversed(retained_energies)
+    ][::-1]
+
+    notes = [str(warning.message) for warning in caught]
+    raw_imaginary_count = int(np.count_nonzero(np.iscomplex(all_energies)))
+    if raw_imaginary_count:
+        notes.append(
+            f"The input spectrum contains {raw_imaginary_count} imaginary modes. "
+            "These thermochemistry values do not establish structural stability; "
+            "inspect the complete spectrum."
+        )
+    if len(atoms) > 1 and not mode_indices:
+        notes.append("No vibrational modes contributed to thermochemistry.")
+    for note in notes:
+        logger.warning(note)
+
+    return {
+        "enthalpy": enthalpy,
+        "entropy": entropy,
+        "gibbs_free_energy": gibbs_free_energy,
+        "unit": "eV",
+        "entropy_unit": "eV/K",
+        "ase_version": ase.__version__,
+        "vib_selection": "highest",
+        "ignore_imag_modes": True,
+        "n_imag": int(thermo.n_imag),
+        "raw_imaginary_mode_count": raw_imaginary_count,
+        "warnings": notes,
+    }, mode_indices
 
 
 def get_symmetry_number(atomsdata: AtomsData) -> int:
@@ -489,7 +594,7 @@ def _energy_result_metadata(
     converged: Optional[bool] = None,
     optimization_steps: Optional[int] = None,
 ) -> dict:
-    """Build consistent energy metadata for a successful tool result."""
+    """Build consistent metadata for a completed energy calculation."""
     result = {
         "driver": driver,
         "potential_energy": potential_energy,
@@ -497,7 +602,7 @@ def _energy_result_metadata(
         "results_file": os.path.abspath(results_file),
     }
     if converged is not None:
-        result["converged"] = converged
+        result["converged"] = bool(converged)
     if optimization_steps is not None:
         result["optimization_steps"] = optimization_steps
     return result
@@ -753,6 +858,7 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
             pbc=atoms.pbc,
         )
         thermo_data: dict = {}
+        thermo_error: Optional[Exception] = None
         vib_data: dict = {}
         ir_data: dict = {}
         ir_plot_path: Optional[str] = None
@@ -810,9 +916,27 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
                 logger.info("Vibrational analysis complete")
 
                 all_energies = vib.get_energies()
-                mode_indices = _vibrational_mode_indices(
-                    final_structure, len(all_energies)
-                )
+                if driver == "thermo":
+                    vib_data.update(
+                        mode_indices=[],
+                        all_modes=[
+                            _vibrational_mode_record(i, e)
+                            for i, e in enumerate(all_energies)
+                        ],
+                    )
+                    try:
+                        thermo_data, mode_indices = _calculate_thermochemistry(
+                            atoms, final_structure, all_energies, potential_energy,
+                            calc_model, temperature, pressure,
+                        )
+                        vib_data["mode_indices"] = mode_indices
+                    except Exception as exc:
+                        logger.exception("Thermochemistry failed; preserving vibration results")
+                        thermo_error, mode_indices = exc, []
+                else:
+                    mode_indices = _vibrational_mode_indices(
+                        final_structure, len(all_energies)
+                    )
 
                 for mode_index in mode_indices:
                     e = all_energies[mode_index]
@@ -916,44 +1040,16 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
                     )
 
         # ---- Thermochemistry ----
-        if driver == "thermo":
-            from ase.thermochemistry import IdealGasThermo
-
-            logger.info("Computing thermochemistry (T=%s K, P=%s Pa)", temperature, pressure)
-            if len(atoms) == 1:
-                geometry, symmetrynumber = "monatomic", 1
-            else:
-                linear = is_linear_molecule(final_structure)
-                geometry = "linear" if linear else "nonlinear"
-                symmetrynumber = get_symmetry_number(final_structure)
-
-            # IdealGasThermo expects total spin S; calculators expose 2S+1.
-            multiplicity = getattr(calc_model, "get_multiplicity", lambda: None)()
-            if multiplicity is None:
-                logger.warning(
-                    "%s does not report spin multiplicity; assuming a singlet "
-                    "(multiplicity=1) for thermochemistry. Electronic-spin entropy "
-                    "is omitted for open-shell species.",
-                    type(calc_model).__name__,
+        if driver == "thermo" and len(atoms) == 1:
+            vib_data.update(mode_indices=[], all_modes=[])
+            try:
+                thermo_data, _ = _calculate_thermochemistry(
+                    atoms, final_structure, [], potential_energy, calc_model,
+                    temperature, pressure,
                 )
-                multiplicity = 1
-            thermo = IdealGasThermo(
-                vib_energies=all_energies,
-                potentialenergy=potential_energy,
-                atoms=atoms,
-                geometry=geometry,
-                symmetrynumber=symmetrynumber,
-                spin=(multiplicity - 1) / 2.0,
-            )
-            enthalpy = float(thermo.get_enthalpy(temperature, verbose=False))
-            entropy = float(thermo.get_entropy(temperature, pressure, verbose=False))
-            thermo_data = {
-                "enthalpy": enthalpy,
-                "entropy": entropy,
-                "gibbs_free_energy": enthalpy - temperature * entropy,
-                "unit": "eV",
-                "entropy_unit": "eV/K",
-            }
+            except Exception as exc:
+                logger.exception("Thermochemistry failed; preserving atomic results")
+                thermo_error = exc
 
         # ---- serialise full output ----
         end_time = time.time()
@@ -967,7 +1063,8 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
             simulation_input=simulation_input,
             vibrational_frequencies=vib_data,
             thermochemistry=thermo_data,
-            success=True,
+            success=thermo_error is None,
+            error=str(thermo_error) if thermo_error is not None else "",
             ir_data=ir_data,
             potential_energy=potential_energy,
             single_point_energy=potential_energy,
@@ -985,6 +1082,16 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
             converged=converged,
             optimization_steps=optimization_steps,
         )
+        if thermo_error is not None:
+            return {
+                "status": "failure",
+                "error_type": type(thermo_error).__name__,
+                "message": (
+                    f"Thermochemistry failed: {thermo_error}. "
+                    f"Completed structure, energy, and vibration results saved to {abs_output}"
+                ),
+                **energy_metadata,
+            }
         if driver == "opt":
             if converged:
                 message = f"Geometry optimization converged. Results saved to {abs_output}"
--- a/src/chemgraph/tools/ase_core.py
+++ b/src/chemgraph/tools/ase_core.py
@@ -734,7 +734,7 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
     atoms.info.update(system_info)
     atoms.calc = calc
 
-    if driver == "ir":
+    if driver == "ir" and len(atoms) > 1:
         # Infrared calls this ASE method at every displacement. Check it before
         # optimization and vibrations; a result-only fallback cannot support IR.
         from ase.calculators.calculator import PropertyNotImplementedError
--- a/src/chemgraph/tools/ase_core.py
+++ b/src/chemgraph/tools/ase_core.py
@@ -876,15 +876,16 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
             mol_stem = (
                 Path(input_structure_file).stem if input_structure_file else "mol"
             )
-            # Remove previous modes even when this run has no vibrations.
             freq_file = Path(_resolve_path(f"frequencies_{mol_stem}.csv"))
-            freq_file.unlink(missing_ok=True)
             traj_dest_dir = _resolve_path("")
             stale_traj_pattern = (
                 glob.escape(_resolve_path(f"{mol_stem}_vib.")) + "*.traj"
             )
-            for stale_traj_file in glob.glob(stale_traj_pattern):
-                os.unlink(stale_traj_file)
+            if len(atoms) == 1:
+                # Single atoms have no replacement vibration artifacts.
+                freq_file.unlink(missing_ok=True)
+                for stale_traj_file in glob.glob(stale_traj_pattern):
+                    os.unlink(stale_traj_file)
 
             if driver == "ir":
                 ir_data = {
@@ -893,12 +894,13 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
                     "spectrum_intensities": [],
                     "spectrum_intensities_units": "D/Å^2 amu^-1",
                 }
-                for name in (
-                    f"ir_spectrum_{mol_stem}.png",
-                    f"ir_spectrum_{mol_stem}.csv",
-                    f"ir_peaks_{mol_stem}.csv",
-                ):
-                    Path(_resolve_path(name)).unlink(missing_ok=True)
+                if len(atoms) == 1:
+                    for name in (
+                        f"ir_spectrum_{mol_stem}.png",
+                        f"ir_spectrum_{mol_stem}.csv",
+                        f"ir_peaks_{mol_stem}.csv",
+                    ):
+                        Path(_resolve_path(name)).unlink(missing_ok=True)
 
         if driver in {"vib", "thermo", "ir"} and len(atoms) > 1:
             logger.info("Starting vibrational analysis (driver=%s)", driver)
@@ -962,6 +964,9 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
 
                 if traj_dest_dir:
                     os.makedirs(traj_dest_dir, exist_ok=True)
+                # Keep previous modes until all replacements have been written.
+                for stale_traj_file in glob.glob(stale_traj_pattern):
+                    os.unlink(stale_traj_file)
                 for mode_index in mode_indices:
                     traj_file = os.path.join(tmpdir, f"vib.{mode_index}.traj")
                     dest_name = f"{mol_stem}_{Path(traj_file).name}"
--- a/src/chemgraph/tools/ase_core.py
+++ b/src/chemgraph/tools/ase_core.py
@@ -269,7 +269,9 @@ def get_symmetry_number(atomsdata: AtomsData) -> int:
         pbc=atomsdata.pbc,
     )
     aaa = AseAtomsAdaptor()
-    molecule = aaa.get_molecule(atoms)
+    # Rotational symmetry depends on geometry, not the calculator's electronic
+    # state. Reconstructed Atoms lack magnetic moments for radical species.
+    molecule = aaa.get_molecule(atoms, charge_spin_check=False)
     pga = PointGroupAnalyzer(molecule)
     return pga.get_rotational_symmetry_number()
 
@@ -751,7 +753,17 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
         # --------------------------------------------------------------
         # Vibrational / thermo / IR analysis
         # --------------------------------------------------------------
+        all_energies = []
         if driver in {"vib", "thermo", "ir"}:
+            vib_data = {
+                "energies": [],
+                "energy_unit": "meV",
+                "frequencies": [],
+                "frequency_unit": "cm-1",
+            }
+        if driver in {"vib", "thermo", "ir"} and not (
+            driver == "thermo" and len(atoms) == 1
+        ):
             logger.info("Starting vibrational analysis (driver=%s)", driver)
             from ase.vibrations import Vibrations
             from ase import units
@@ -769,13 +781,6 @@ def _run_ase_core(params: ASEInputSchema) -> dict:
                 vib.clean()
                 vib.run()
                 logger.info("Vibrational analysis complete")
-
-                vib_data = {
-                    "energies": [],
-                    "energy_unit": "meV",
-                    "frequencies": [],
-                    "frequency_unit": "cm-1",
-                }
 
                 all_energies = vib.get_energies()
                 mode_indices = _vibrational_mode_indices(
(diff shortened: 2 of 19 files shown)


src/chemgraph/tools/ase_core.py
```diff
@@ -861,55 +866,39 @@ def run_ase_core(params: ASEInputSchema) -> dict:
...
+            thermo = IdealGasThermo(
+                vib_energies=all_energies,
+                potentialenergy=potential_energy,
+                atoms=atoms,
+                geometry=geometry,
+                symmetrynumber=symmetrynumber,
+                spin=(multiplicity - 1) / 2.0,
+            )
+            thermo_data = {
+                "enthalpy": float(thermo.get_enthalpy(temperature=temperature)),
```
keceli, line 892: **Regression.** `temperature` is `Optional[float] = Field(default=None)` (`ase_input.py:326`) with no validator requiring it for `driver="thermo"`. The monatomic branch you removed was the only thermo path that never read it — on main it returned `potential_energy` and `0.0` directly (`ase_core.py:867-873`). Now every monatomic run reaches ASE, which formats `'... T = %.2f K:' % temperature` and raises.

```python
run_ase_core(ASEInputSchema(input_structure_file='cu.xyz',
             calculator={'calculator_type': 'emt'}, driver='thermo'))

# main: {'status': 'success', 'thermochemistry': {'enthalpy': 3.51, ...}}
# here: {'status': 'failure', 'error_type': 'TypeError',
#        'message': 'must be real number, not NoneType'}
```

That shape — an atomic reference with no explicit temperature — is exactly what an agent computing an atomization energy produces, and the schema gives the model no default to anchor on. All 17 new tests pass an explicit temperature, so none cover it.

A `298.15` default on the field, or a `model_validator` requiring temperature when `driver == "thermo"`, would close it.

**Separately, on these same three getters:** ASE defaults them to `verbose=True`, writing ~2 KB / 53 formatted lines to raw stdout per call (`get_gibbs_energy` re-invokes the other two internally, so enthalpy and entropy each print and compute twice). `ThermoChem._vprint` is a bare `sys.stdout.write`, which the stderr logging setup doesn't capture. `mcp_tools.py` guards its calls with `redirect_stdout`, but `mcp/ase_mcp_hpc.py` calls `run_ase_core` three times with no guard, and `server_utils.py:31` defaults to `--transport stdio` — where stdout *is* the JSON-RPC channel. This is pre-existing for polyatomic thermo, but the diff newly routes monatomic runs, previously silent, through it. `verbose=False` on all three plus `G = H - T*S` fixes both the noise and the double computation.

src/chemgraph/tools/ase_core.py
```diff
@@ -861,55 +866,39 @@ def run_ase_core(params: ASEInputSchema) -> dict:
...
+                geometry, symmetrynumber = "monatomic", 1
+            else:
+                linear = is_linear_molecule(final_structure)
+                geometry = "linear" if linear else "nonlinear"
+                symmetrynumber = get_symmetry_number(final_structure)
+
+            # IdealGasThermo expects total spin S; calculators expose 2S+1.
+            multiplicity = getattr(calc_model, "get_multiplicity", lambda: None)() or 1
+            thermo = IdealGasThermo(
+                vib_energies=all_energies,
```
keceli, line 884: This hands `IdealGasThermo` the full 3N complex array rather than the `mode_indices` subset the same function computed 135 lines above.

`_vibrational_mode_indices` selects `range(6, 3N)` in eigenvalue order. ASE 3.25 instead does `vib_energies.sort(key=np.abs)` then takes `[-(3N-6):]` — magnitude order — and then `_clean_vib_energies(ignore_imag_modes=False)` raises on anything imaginary. The two selections disagree whenever a genuine imaginary mode is smaller in magnitude than a trans/rot residual.

Reproduced: square-planar Cu4 + EMT + `driver='thermo'` returns `{'status':'failure','error_type':'ValueError','message':'Imaginary vibrational energies are present.'}`, discarding the whole run including the optimized structure and the frequencies that would otherwise have been written. When the imaginary mode is the smaller one it silently succeeds but integrates a different mode set than the one reported in `vibrational_frequencies`.

Pre-existing verbatim, but this diff rewrote and relocated the call, and the correct subset is already in scope:

```python
vib_energies=[all_energies[i] for i in mode_indices],
```

src/chemgraph/tools/ase_core.py
```diff
@@ -861,55 +866,39 @@ def run_ase_core(params: ASEInputSchema) -> dict:
...
+            logger.info("Computing thermochemistry (T=%s K, P=%s Pa)", temperature, pressure)
+            if len(atoms) == 1:
+                geometry, symmetrynumber = "monatomic", 1
+            else:
+                linear = is_linear_molecule(final_structure)
+                geometry = "linear" if linear else "nonlinear"
+                symmetrynumber = get_symmetry_number(final_structure)
+
+            # IdealGasThermo expects total spin S; calculators expose 2S+1.
+            multiplicity = getattr(calc_model, "get_multiplicity", lambda: None)() or 1
```
keceli, line 882: `or 1` silently forces a singlet, so the open-shell atoms and radicals this PR targets get zero electronic entropy on most calculators.

Only mace/tblite/orca/nwchem/fairchem define `get_multiplicity`. `EMTCalc` and `AIMNET2Calc` have no multiplicity field at all, and `mace_parsl_schema.mace_input_schema` has neither multiplicity nor charge (`parsl_tools.py:61` hardcodes `calculator_type='mace_mp'`), so the entire MACE parsl/MCP surface can never set it.

Measured: a Cu atom at 298.15 K gives S = 0.00166373 eV/K = 160.5 J/(mol K). Literature Cu(g) is 166.4; adding the omitted `kB*ln 2` gives 166.29. Per-atom error in G runs 0.018 eV (H, Cl, Cu doublets) to 0.036 eV (N quartet) — one-directional, so it accumulates in atomization and BDE workflows with no warning.

Worth either defaulting from the element's ground-state term symbol, or logging when a thermo run falls back to the singlet so the omission is visible.

There's also a consistency hazard one layer down, now that the new docs paragraph formalizes this coupling: `nwchem_calc.py:116` reports the multiplicity here even when `get_calculator` drops it from the input deck (`theory='scf'`, or any theory outside `{dft,mp2,ccsd,tce,tddft}`), and `mace_calc.py:211` returns `{}` from `get_atoms_properties()` for every type except `mace_polar`. Both pair a spin-agnostic enthalpy with a spin-shifted entropy.

src/chemgraph/tools/ase_core.py
```diff
@@ -714,7 +716,17 @@ def run_ase_core(params: ASEInputSchema) -> dict:
...
         # --------------------------------------------------------------
+        all_energies = []
         if driver in {"vib", "thermo", "ir"}:
+            vib_data = {
+                "energies": [],
+                "energy_unit": "meV",
+                "frequencies": [],
+                "frequency_unit": "cm-1",
+            }
+        if driver in {"vib", "thermo", "ir"} and not (
```
keceli, line 727: Two things about this guard.

**It special-cases one driver instead of the physical fact.** `driver="vib"` and `driver="ir"` on a single atom still run the full finite-difference loop. Verified on a Cu atom: `vib` returns success with `energies: []` after 7 calculator evaluations that `_vibrational_mode_indices` then discards at line 243 (`if num_atoms == 1: return []`), and `ir` fails with `PropertyNotImplementedError: dipole property not implemented` after another 7. With ORCA/NWChem that's 14 wasted SCFs or subprocess launches. `docs/calculators.md:42` states single atoms skip finite-difference vibrations, which only holds for thermo. The general condition is `len(atoms) > 1`, the same shape already used for the optimizer at line 684.

**Skipping the block also skips its cleanup.** `freq_file.unlink()` and the rewrite at 764-770, and `for f in glob.glob(f'{mol_stem}_vib.*.traj'): os.unlink(f)` at 783-787, live inside this block and nowhere else.

Reproduced: run thermo on a Cu2 dimer via `mol.xyz` in a log dir, overwrite `mol.xyz` with a single atom, run thermo again — `frequencies_mol.csv` and `mol_vib.5.traj` from the dimer are still there, where main unlinked both. `main_interface.py:1602` falls back to `_latest_artifact_path(log_dir, 'frequencies*.csv')` and `artifacts.py:126` buckets `*_vib.*.traj`, so the artifact panel shows the previous molecule's modes beside the atom's result. Hoisting the cleanup above the guard would fix it.

src/chemgraph/tools/ase_core.py
```diff
@@ -269,7 +269,9 @@ def get_symmetry_number(atomsdata: AtomsData) -> int:
         pbc=atomsdata.pbc,
     )
     aaa = AseAtomsAdaptor()
-    molecule = aaa.get_molecule(atoms)
+    # Rotational symmetry depends on geometry, not the calculator's electronic
+    # state. Reconstructed Atoms lack magnetic moments for radical species.
+    molecule = aaa.get_molecule(atoms, charge_spin_check=False)
```
keceli, line 274: The fix is correct — on pymatgen 2025.3.10, OH/NO/CH3 raise `ValueError: Charge of 0 and spin multiplicity of 1 is not possible` without this flag and return sigma 1/1/6 with it.

Two durability notes on the same function:

**`charge_spin_check` isn't a declared parameter of `get_molecule`.** The signature is `get_molecule(atoms, cls=Molecule, **cls_kwargs)`, so this rides undeclared kwargs through into `Molecule.__init__`. Nothing at the call site validates it, and `pyproject.toml` pins a bare `"pymatgen"` with no floor — so a release that stops forwarding unknown kwargs turns every polyatomic `driver='thermo'` run and every `get_symmetry_number` tool call into a `TypeError` several frames deep in pymatgen. A `pymatgen>=` floor matching `environment.yml`, or constructing the `Molecule` directly, would pin this down.

**Monatomic input raises here.** `get_symmetry_number(AtomsData(numbers=[8], positions=[[0,0,0]]))` raises `AttributeError: 'PointGroupAnalyzer' object has no attribute 'symmops'`. `run_ase_core` now avoids it via the `geometry, symmetrynumber = "monatomic", 1` branch, but this function is also exported as a standalone agent tool (`registry/tools.py:133` -> `ase_tools.py:118`), so an agent reasoning about an atomic reference gets an unhandled `AttributeError` instead of the obvious `1`. The new `_MOLECULES` table has no single-atom case, and `tests/test_tools.py:106` only asserts `isinstance(..., int)`.

Minor, same function: `cell` and `pbc` are forwarded into the rebuilt `Atoms`, but `get_molecule` discards both — `Molecule` takes no `Lattice`. An H2O whose second H sits at `x = 10 - 0.76` in a 10 A periodic cell returns sigma = 1 instead of 2, which is `kB*ln 2` = 0.0178 eV in T*S at 298 K. Since this is an exported tool, an agent handing it a trajectory frame gets a confidently wrong integer rather than an error.

src/chemgraph/tools/ase_core.py
```diff
@@ -861,55 +866,39 @@ def run_ase_core(params: ASEInputSchema) -> dict:
...
+            thermo_data = {
+                "enthalpy": float(thermo.get_enthalpy(temperature=temperature)),
+                "entropy": float(
+                    thermo.get_entropy(temperature=temperature, pressure=pressure)
+                ),
+                "gibbs_free_energy": float(
+                    thermo.get_gibbs_energy(temperature=temperature, pressure=pressure)
+                ),
+                "unit": "eV",
+                "entropy_unit": "eV/K",
```
keceli, line 900: `entropy_unit` is write-only — this is the only place it's set, and nothing in `src/` reads it.

`report_tools.py:630` emits the literal `<span class="entropy-unit">eV/K</span>` and line 802 multiplies by `EV_TO_KJMOL` unconditionally, both independent of what the payload declares. A producer that ever emits `entropy_unit: 'J/(mol K)'` gets silently relabeled eV/K and scaled by 96.485 — five orders of magnitude off under a confident label.

It also makes the new `@pytest.mark.parametrize("legacy", [False, True])` at `tests/test_report.py:170` vacuous: both variants generate byte-identical HTML, so a test named "distinguishes entropy units" varies nothing the renderer reads.

Either have the renderer use `thermochemistry.get('entropy_unit', 'eV/K')` and derive the factor from it, or drop the field and keep the unit purely a renderer concern.

src/chemgraph/tools/report_tools.py
```diff
@@ -621,17 +621,18 @@ def add_additional_info_to_html(html_content: str, ase_output: ASEOutputSchema)
...
-                f'<div><strong>Enthalpy:</strong> <span class="energy-value" data-ev="{enthalpy_ev}">{enthalpy_ev:.6f}</span></div>'
+                f'<div><strong>Enthalpy:</strong> <span class="energy-value" data-ev="{enthalpy_ev}">{enthalpy_ev:.6f}</span> <span class="energy-unit">eV</span></div>'
             )
         if "entropy" in ase_output.thermochemistry:
-            entropy_ev = ase_output.thermochemistry['entropy']
+            # Current and legacy ASE results express entropy in eV/K.
+            entropy_ev_k = ase_output.thermochemistry['entropy']
             thermo_info.append(
-                f'<div><strong>Entropy:</strong> <span class="energy-value" data-ev="{entropy_ev}">{entropy_ev:.6f}</span></div>'
+                f'<div><strong>Entropy:</strong> <span class="entropy-value" data-ev-k="{entropy_ev_k}">{entropy_ev_k:.6f}</span> <span class="entropy-unit">eV/K</span></div>'
```
keceli, line 630: Renaming the span from `energy-value` to `entropy-value` drops it out of the only rule that sizes these numbers.

The stylesheet defines `.energy-value { display: inline-block; min-width: 100px; }` at line 171 and no `.entropy-value` rule at all (`'.entropy-value {' in html` is False on a generated report). Enthalpy and Gibbs keep the reserved 100px column; Entropy, sitting between them, collapses to natural width, so the unit labels no longer line up and the column jitters on every unit toggle.

Adding `.entropy-value` to the selector at line 171 restores it.

src/chemgraph/tools/report_tools.py
```diff
@@ -791,6 +792,15 @@ def add_additional_info_to_html(html_content: str, ase_output: ASEOutputSchema)
                     label.textContent = unit === 'ev' ? 'eV' : 
                                      unit === 'kjmol' ? 'kJ/mol' : 'kcal/mol';
                 });
+                document.querySelectorAll('.entropy-unit').forEach(label => {
+                    label.textContent = unit === 'ev' ? 'eV/K' :
+                                     unit === 'kjmol' ? 'kJ/(mol K)' : 'kcal/(mol K)';
+                });
+                document.querySelectorAll('.entropy-value').forEach(cell => {
```
keceli, line 799: This branch hardcodes `.toFixed(6)` while the `.energy-value` loop it was copied from sets `precision = 2` for kJ/mol and kcal/mol (lines 816/819).

So one click on `kJ/mol` renders the three rows as `-1264.63`, `0.188869`, `-1320.93` in the same block — entropy at six decimals beside energies at two.

Entropy genuinely needs more digits than energy here (it's ~1e-3 eV/K), so matching `precision` exactly would lose information; something like 6 for eV/K and 4 for the molar units would keep both readable.

src/chemgraph/tools/report_tools.py
```diff
@@ -643,7 +644,7 @@ def add_additional_info_to_html(html_content: str, ase_output: ASEOutputSchema)
                     <button onclick="toggleEnergyUnit('kjmol')" data-unit="kjmol">kJ/mol</button>
                     <button onclick="toggleEnergyUnit('kcalmol')" data-unit="kcalmol">kcal/mol</button>
                 </div>
-                <strong>Thermochemistry Values</strong> (<span class="energy-unit">eV</span>):<br>
+                <strong>Thermochemistry Values</strong>:<br>
```
keceli, line 647: The toggle is still labelled "Energy Unit:" with `eV` / `kJ/mol` / `kcal/mol` buttons, but it now switches entropy too — clicking `kJ/mol` relabels the entropy row to `kJ/(mol K)`.

Button text and produced value disagree, which is the ambiguity this PR set out to remove. Someone clicking `kcal/mol` to read enthalpy silently rewrites entropy as well; screenshotting the panel while remembering only the button pressed carries off an entropy in the wrong unit.

`docs/calculators.md:41` compounds it — "HTML reports also support kJ/(mol K) and kcal/(mol K) for entropy" sends a reader looking for a separate selector that was never added. The doc sentence and this label are the only two places the unit control is described, and they disagree.

Relabelling to just "Units:" would make it accurate.

Also, now that the block-header `.energy-unit` span is gone: a payload carrying entropy but neither enthalpy nor Gibbs renders the full toggle with no unit label anywhere.

tests/test_thermochemistry.py
```diff
@@ -0,0 +1,136 @@
...
+
+
+@pytest.mark.parametrize("temperature,pressure", [(250, 101325), (500, 101325), (500, 202650)])
+@pytest.mark.parametrize("multiplicity", [1, 2, 3])
+def test_monatomic_thermo_matches_ase_without_vibrations(
+    tmp_path, monkeypatch, temperature, pressure, multiplicity,
+):
+    vibrations = Mock(side_effect=AssertionError("an atom has no vibrational modes"))
+    monkeypatch.setattr(ase.vibrations, "Vibrations", vibrations)
+    monkeypatch.setattr(EMTCalc, "get_multiplicity", lambda self: multiplicity, raising=False)
```
keceli, line 53: `raising=False` here fabricates a method that doesn't exist — `hasattr(EMTCalc, 'get_multiplicity')` is False.

So six of the nine parametrized cases exercise a spin EMT can never report. Renaming `get_multiplicity` across the five calculator schemas that genuinely implement it, or a typo in the `getattr` key at `ase_core.py:882`, leaves all nine green while every real EMT run silently falls back to the singlet default.

`MaceCalc(multiplicity=...)` actually implements it, and the molecular test in this same file already uses that — worth doing the same here so the test exercises the real lookup path.

Two more assertions in this test are weaker than they read:

- `vibrations.assert_not_called()` (line 70) and `assert not list(tmp_path.glob("*.traj"))` (line 71) can't fail in the regression they guard. Run this file against main and it dies earlier, at line 32 — `run_ase_core`'s blanket `except Exception` swallows the Mock's `AssertionError` and returns `status='failure'`. Even if reached, main produced zero `.traj` for a monatomic thermo run; its only stale artifact was `frequencies_cu.csv`, which a `*.csv` glob would catch and this one doesn't.
- `result["result"]["thermochemistry"] == output["thermochemistry"]` (line 34) compares the same dict round-tripped through exact JSON float serialization, so it can only trip on NaN.

The `_assert_reference` comparison and the analytic `potential_energy + 2.5 * kB * T` check are both real, so the test does its main job — these are the surrounding assertions.

tests/test_thermochemistry.py
```diff
@@ -0,0 +1,136 @@
...
+    atoms = Atoms(symbols, positions=positions)
+    assert get_symmetry_number(AtomsData(numbers=atoms.numbers, positions=positions)) == symmetry
+
+
+@pytest.mark.parametrize("symbols,positions,symmetry,multiplicity,geometry", _MOLECULES)
+def test_molecular_thermo_retains_spin_and_vibrations(
+    tmp_path, monkeypatch, symbols, positions, symmetry, multiplicity, geometry,
+):
+    atoms = Atoms(symbols, positions=positions)
+    energies = np.zeros(3 * len(atoms))
```
keceli, line 93: This fixture can't catch the mode-selection bug it looks like it covers.

`np.zeros(3 * len(atoms))` with only the last `mode_count` entries nonzero is already in `sort(key=np.abs)` order and contains no imaginary values. Real `Vibrations.get_energies()` returns `dtype complex128` with the near-zero trans/rot modes stored as pure imaginary (verified on EMT Cu2) — which is precisely what makes ASE's magnitude-based reselection diverge from `mode_indices`.

Line 128 then builds the reference `IdealGasThermo(vib_energies=energies, ...)` from that same array, so ASE applies identical slicing to both sides. Changing `vib_energies=all_energies` to `[all_energies[i] for i in mode_indices]` — or the reverse — leaves all 17 tests green.

A fixture returning complex energies with imaginary trans/rot entries would make this test discriminating.

Also `mode_count = 1 if geometry == "linear" else 3` hardcodes the current atom counts rather than `3N-5` / `3N-6`, so adding CO2 to `_MOLECULES` silently breaks the arithmetic.

docs/calculators.md
```diff
@@ -36,6 +36,13 @@ Calculator-backed tools cover operations such as:
 Support depends on the selected calculator. A valid property for one engine may
 not exist for another.
 
+Ideal-gas thermochemistry uses the requested temperature and pressure. Enthalpy
+and Gibbs energy are reported in eV; entropy is reported in eV/K with a separate
+`entropy_unit` field. HTML reports also support kJ/(mol K) and kcal/(mol K) for
+entropy. Single atoms include translational and electronic-spin contributions
```
keceli, line 42: Two claims in this paragraph don't hold as written.

"Single atoms include translational and electronic-spin contributions **without running finite-difference vibrations**" is true only for `driver="thermo"`. The guard at `ase_core.py:727` is `driver == "thermo" and len(atoms) == 1`, so `vib` and `ir` on a single atom still run the full displacement loop (see that comment).

"**Calculator multiplicity determines** the electronic-spin contribution" is unreachable on EMT, AIMNet2, and the entire MACE parsl/MCP surface — none of them can report a multiplicity, so those runs always get S = 0 regardless of the species. Worth saying that the contribution is included *when the calculator reports a multiplicity*, and that it defaults to a singlet otherwise, so readers computing atomization energies with MACE know the term is missing.

keceli (changes requested):
The physics here is right and the diagnosis is correct. Monatomic thermo returning `entropy = 0.0` with `H = G = E_pot` at every T and P was straightforwardly wrong, and routing it through `IdealGasThermo` with `geometry="monatomic"` gives the proper Sackur-Tetrode translational term. The `charge_spin_check=False` reasoning is right too — `PointGroupAnalyzer` reads geometry only, so the check was rejecting OH and NO over a property it doesn't use.

I checked the restructuring specifically, since dedenting thermo out of the vibration block is where this could have broken: `all_energies` is initialized before either branch, `final_structure` is in scope from line 706, the relocated block is still inside the outer `try` ending at 984, and `vibrational_frequencies` output is byte-identical to main. 15 of the 17 new tests fail on main, so the suite does gate the fix.

Requesting changes on one regression, plus a set of things worth folding in while this is open.

### The blocker

`temperature` is `Optional[float] = None` with no validator, and the monatomic short-circuit you removed was **the only thermo path that never read it**. Details inline at `ase_core.py:892` — short version, `driver="thermo"` on a single atom with no temperature returns `status: "success"` on main and `TypeError: must be real number, not NoneType` on this branch. That's an agent pulling an atomic reference for an atomization energy, which is the workflow this fix exists to enable, and `temperature` is the field an LLM is most likely to omit since the schema gives it no default. All 17 new tests pass an explicit temperature.

### Already ruled out, so you don't re-spend the time

- `data-ev-k` -> `cell.dataset.evK` is the correct camelCase mapping, and the converter reads from the dataset so toggling is idempotent.
- `symmetrynumber=1` for monatomic is required, not merely harmless — `get_entropy` raises if `sigma` or `spin` is None.
- `Vibrations.run()` and `write_mode()` leave `atoms.positions` untouched, so the relocated thermo block sees the optimized geometry.
- `get_symmetry_number` returns the correct sigma for every non-monatomic species tested.
- `run_ase_core` is the only thermochemistry producer; parsl/MACE/academy all delegate to it.

One environment note unrelated to the diff: `ruff check .` passes on the long new lines only because `E501` isn't in the selected rule set (`select = ["E4","E7","E9","F"]`); `ruff format --check` would reformat them.

---

🤖 Findings generated with [Claude Code](https://claude.com/claude-code) (Opus 5) and posted by @keceli.

src/chemgraph/tools/ase_core.py
tdpham2: Fixed in [3e5b0b37b](https://github.com/argonne-lcf/ChemGraph/commit/3e5b0b37b). ASE single and ensemble inputs now default to 298.15 K, including explicit null input, reject nonpositive/nonfinite temperatures, and persist the effective temperature. The regressions cover omitted, null, explicit, and invalid values, including a real EMT atomic thermo run with no supplied temperature.

The enthalpy and entropy getters now use `verbose=False`; Gibbs energy is calculated as `H - T*S` from those results. Atomic tests assert empty stdout, and molecular tests check that the thermochemistry tables are absent.

Validation for the revision: `ruff check .` and `git diff --check` passed; the full suite with Academy/Parsl/Globus dependencies passed with **858 passed, 21 skipped, 2 deselected**. Live-service tests were not enabled.

src/chemgraph/tools/ase_core.py
tdpham2: I reproduced the selection discrepancy, but the proposed slice needs a separate scientific fix. The square-planar Cu4/EMT probe contained imaginary entries around 0.00748031i and 0.00398170i eV. Selecting `energies[6:]` dropped both and accepted a spectrum containing small real residuals instead. We need to distinguish rigid-body residuals from unstable molecular modes before treating that as a correction.

There is also a version dependency: installed ASE 3.29 defaults to `vib_selection='highest'`; its `abs_highest` option reproduces the magnitude-selection behavior described here. I checked both policies, rather than running a separate ASE 3.25 installation. [ASE documentation](https://docs.ase-lib.org/ase/thermochemistry/thermochemistry.html).

Production mode selection is unchanged in this revision. I am leaving this thread open so we can settle mode identification, imaginary-mode handling, consistency with reported frequencies, and preservation of partial results together.

src/chemgraph/tools/ase_core.py
tdpham2: Implemented the warning option in [3e5b0b37b](https://github.com/argonne-lcf/ChemGraph/commit/3e5b0b37b). When the multiplicity accessor is absent or returns None, thermochemistry logs that it is assuming a singlet and that open-shell electronic-spin entropy is omitted. The docs now state this limitation and distinguish the entropy correction from the calculator's potential-energy capabilities.

Tests cover the actual EMT fallback, a real NWChem schema whose accessor returns None with an EMT engine substituted, and explicit multiplicities through the real MACE accessor.

Automatic ground-state inference, new Parsl spin inputs, and calculator-specific electronic-state propagation remain outside this revision. I am leaving the thread open because the broader consistency concerns still need follow-up.

src/chemgraph/tools/ase_core.py
tdpham2: Fixed in [3e5b0b37b](https://github.com/argonne-lcf/ChemGraph/commit/3e5b0b37b). The guard now uses the atom count for all three drivers: `thermo`, `vib`, and `ir`. Single atoms skip both displacement engines, retain empty vibration arrays with units, and return empty IR spectrum arrays with units when requested.

Same-stem frequency CSV and mode-trajectory cleanup now runs before the guard. IR requests also remove stale spectrum PNG/CSV and peak CSV files. Atomic IR completion messages explain that no spectrum exists.

The regression seeds previous artifacts, verifies their removal and preservation of unrelated files, and asserts neither `Vibrations` nor `Infrared` is constructed. It also covers filenames containing brackets so glob characters cannot select another molecule's modes.

src/chemgraph/tools/ase_core.py
tdpham2: The atomic-input crash is fixed in [3e5b0b37b](https://github.com/argonne-lcf/ChemGraph/commit/3e5b0b37b): `get_symmetry_number` returns 1 directly for a single atom, with H/O/Cu coverage alongside the radical and molecular controls.

I retained the adaptor call because forwarding `**cls_kwargs` to the molecule constructor is explicitly documented by [pymatgen](https://pymatgen.org/pymatgen.io.html#pymatgen.io.ase.AseAtomsAdaptor.get_molecule). A minimum version would not guard against a future incompatible release.

The function and calculator docs now specify isolated, unwrapped molecular coordinates. Periodic image reconstruction remains deferred because it needs defined behavior for both wrapped molecules and extended systems. I am leaving this thread open for that remaining concern.

src/chemgraph/tools/ase_core.py
tdpham2: Fixed in [339c57cf3](https://github.com/argonne-lcf/ChemGraph/commit/339c57cf3) by making the canonical-unit contract explicit. The renderer now reads `entropy_unit`: absent metadata means legacy eV/K, explicit eV/K is accepted, and other declared units produce a clear error asking the caller to convert entropy before generating the report.

The error occurs before writing or replacing HTML, so unsupported units cannot be silently relabeled. Tests cover current and legacy inputs, J/(mol K), kJ/(mol K), kcal/(mol K), null, and empty metadata; they also verify that an existing report is preserved. The producer retains `entropy_unit` as meaningful metadata.

src/chemgraph/tools/report_tools.py
tdpham2: Fixed in [339c57cf3](https://github.com/argonne-lcf/ChemGraph/commit/339c57cf3): `.energy-value` and `.entropy-value` now share the existing `display: inline-block; min-width: 100px` rule. Entropy reserves the same number width through unit changes. I verified the shared CSS rule in source; visual browser verification was unavailable in this session.

src/chemgraph/tools/report_tools.py
tdpham2: Applied the suggested precision in [339c57cf3](https://github.com/argonne-lcf/ChemGraph/commit/339c57cf3): six decimals for eV/K and four for kJ/(mol K) and kcal/(mol K). Energy retains its existing precision.

The new `test_report_unit_conversion_javascript` executes the generated converter in QuickJS and checks values, labels, precision, and repeated round trips. It ran successfully here; the test skips when the optional QuickJS runtime is absent.

src/chemgraph/tools/report_tools.py
tdpham2: Renamed both shared controls to “Units:” in [339c57cf3](https://github.com/argonne-lcf/ChemGraph/commit/339c57cf3). The docs now explain that selecting a molar energy unit also changes entropy to the corresponding per-kelvin unit.

I could not reproduce the missing label for entropy-only input: the reviewed head already rendered an `entropy-unit` span containing eV/K. Added explicit entropy-only cases for both current and legacy payloads to preserve that behavior.

tests/test_thermochemistry.py
tdpham2: Updated in [3e5b0b37b](https://github.com/argonne-lcf/ChemGraph/commit/3e5b0b37b). The atomic multiplicity tests now use `MaceCalc(calculator_type='mace_polar', multiplicity=...)` and its real accessor, with engine creation mocked to EMT. A separate test exercises the actual EMT singlet fallback.

I retained the success/no-vibration checks: an unexpected vibration call becoming a failure still fails the test, even if it happens before `assert_not_called()`. Added explicit frequency-CSV checks and seeded stale-artifact cases to cover cleanup directly. Persisted-output consistency remains alongside the independent ASE and analytic enthalpy checks.

The new thermochemistry test file is also formatted with Ruff.

tests/test_thermochemistry.py
tdpham2: Strengthened the stable-molecule fixture in [3e5b0b37b](https://github.com/argonne-lcf/ChemGraph/commit/3e5b0b37b): it now returns complex energies with small imaginary rigid-body residuals, and the reference receives only the fixture's known positive vibrational energies. Mode counts use `3N-5` or `3N-6`, with CO2 added to exercise a linear triatomic.

This still does not discriminate between the disputed selection policies when their selected modes differ. That case needs to accompany the scientific mode-selection fix discussed in [the related thread](https://github.com/argonne-lcf/ChemGraph/pull/233#discussion_r3943243659), where slicing can discard imaginary entries. I am leaving this thread open for that remaining coverage.

docs/calculators.md
tdpham2: Updated in [339c57cf3](https://github.com/argonne-lcf/ChemGraph/commit/339c57cf3). The implementation now skips finite differences for single atoms under `thermo`, `vib`, and `ir`, and the docs name those drivers explicitly.

The multiplicity wording now says that spin entropy uses a reported multiplicity, with a warned singlet fallback when it is unavailable. It also explains that this does not infer a ground-state multiplicity or add spin dependence to the calculator's potential energy.

tests/test_report.py
```diff
@@ -166,6 +167,120 @@
...
+            cell.parentElement = {querySelector: () => ({textContent: ''})};
+        });
+        const document = {
+            querySelectorAll: selector => elements[selector] || [],
+            addEventListener: () => {},
+        };
+    """)
+    script = next(
+        script
+        for script in re.findall(r'<script>(.*?)</script>', content, re.S)
```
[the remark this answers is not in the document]
tdpham2: Fixed in [e0c40f1de](https://github.com/argonne-lcf/ChemGraph/commit/e0c40f1de). The JavaScript regression now extracts script contents with Python’s `HTMLParser` instead of an HTML-matching regex. The test covers lowercase, uppercase, and mixed-case script tags with attributes, and executes the extracted converter through the existing unit-conversion checks.

Validation: all 18 report tests passed; the full suite passed with **860 passed, 21 skipped, 2 deselected**. Ruff and whitespace checks passed. Python CodeQL analysis on the pushed commit reports zero findings, and GitHub marks this alert as fixed.

src/chemgraph/tools/ase_core.py
tdpham2: Implemented the agreed ASE 3.29 policy in 4f1e239b4, 7d9e6441e, and 05984ee6a.

The full 3N input is now intentional: IdealGasThermo receives explicit `vib_selection="highest", ignore_imag_modes=True`, and ChemGraph derives the reported thermochemistry frequencies, CSV entries, and trajectory indices from the energies ASE actually retained. Duplicate energies at the selection boundary retain their original ASE indices. ChemGraph does not apply a separate imaginary-frequency filter.

The complete input spectrum is preserved in `vibrational_frequencies.all_modes`. Output records the ASE version, policy, cleanup count, raw imaginary-mode count, and warnings. This distinction matters for the recorded Cu4 probe: its five imaginary input modes are excluded during selection, so ASE reports `n_imag=0`. The report now shows that distinction and states that these values do not establish structural stability.

If thermochemistry raises or yields non-finite H/S/G, the result JSON preserves the completed structure, energy, convergence state, and spectrum; the failure response includes `results_file`.

Validation with ASE 3.29.0: 80 focused thermochemistry/report tests and the full suite (874 passed, 21 skipped, 2 deselected). Both Dockerfiles now pin 3.29.0.

Leaving this thread open for review of the chosen scientific policy.

tests/test_thermochemistry.py
tdpham2: Added discriminating complex-spectrum regressions in 7d9e6441e. Cases include the recorded Cu4/EMT spectrum, imaginary modes remaining after initial selection, selected zero energies, complete removal, duplicate energies at the selection boundary, and nontrivial mode ordering.

The tests assert explicit expected original mode indices and independently compare retained energies and H/S/G with ASE 3.29 configured with `vib_selection="highest", ignore_imag_modes=True`. They also verify that CSV rows and written trajectory IDs match those exact indices and that excluded complex modes remain in the complete spectrum. The linear/nonlinear molecule tests retain the generic 3N-5 / 3N-6 counts and CO2 coverage.

Validation: 80 focused tests passed; full suite 874 passed, 21 skipped, 2 deselected. This addresses the fixture coverage request. The separate mode-selection policy thread remains open.

src/chemgraph/tools/ase_core.py
tdpham2: Follow-up in 3ce56600e: removed ChemGraph's blanket warning for imaginary modes anywhere in the input spectrum. With ASE's `highest` selection, input modes may already be excluded before cleanup; a nonzero raw imaginary-mode count alone no longer produces a warning.

The complete spectrum and `raw_imaginary_mode_count` remain available. We still preserve ASE's cleanup warnings and warn if no vibrational modes contribute. The ASE 3.29 selection/cleanup policy and the modes used for thermochemistry are unchanged; the helper docstrings and calculator documentation now clarify these distinctions.

Recomputing thermochemistry from the saved methane 500 K spectrum now gives `n_imag=0`, `raw_imaginary_mode_count=3`, and `warnings=[]`, with the same nine retained modes and H/S/G (verified to absolute tolerance 1e-12).

Validation: Ruff passed; 80 focused tests passed; full suite: 874 passed, 21 skipped, 2 deselected. Regression tests check both warning metadata and log output, plus HTML rendering with no warning.

keceli: Answering the policy question you left open, since this is the one thing I'd hold the PR for.

**The `highest` + `ignore_imag_modes=True` combination converts a loud failure into a silent wrong answer, and I don't think that trade is right here.**

You were right to push back on my `energies[6:]` suggestion — it did drop genuine imaginary modes and keep small real residuals, and deriving the reported modes from what ASE actually retained is a real improvement. But the selection policy you settled on has the same failure shape, and it's now invisible.

ASE 3.29 sorts with `vib_energies.sort(key=lambda f: (f ** 2).real)`. A purely imaginary mode `0+bj` has key `-b**2` — the *most negative* value in the array — so it sorts to the front and is removed by the `vib_energies[-num_vibs:]` cut **before `_clean_vib_energies` ever runs**. `n_imag` therefore stays 0, no UserWarning is emitted, and `warnings` is empty.

Your own `recorded-cu4` fixture encodes exactly this:

```python
[0.00748031j, 0.00398170j, 3.96854e-8j, 3.96839e-8j, 3.12353e-10j,
 8.62908e-5, 9.16070e-5, 9.16070e-5, 0.0234627, 0.0297504, 0.0297504, 0.0352987]
# expected_indices = [6, 7, 8, 9, 10, 11], cleanup_count = 0
```

`0.00748031j` is ~60 cm-1 imaginary and `0.00398170j` is ~32 cm-1 imaginary. Both disappear via *selection*, so `cleanup_count` is 0 and nothing warns. Meanwhile indices 6 and 7 are `9.16070e-5` eV = **0.74 cm-1** — rigid-body residues promoted into the retained vibrational set.

That substitution is not cosmetic. At E/kT = 0.00356, each of those modes contributes roughly

```
S ≈ k[ (E/kT)/(e^(E/kT) - 1) - ln(1 - e^(-E/kT)) ] ≈ 6.6 k ≈ 5.7e-4 eV/K
```

so the pair adds ~1.14e-3 eV/K, i.e. **T*S ≈ 0.34 eV (~33 kJ/mol)** of numerical noise in the reported Gibbs energy — reported as `status: success`, `n_imag: 0`, `warnings: []`.

On main this input raised `ValueError: Imaginary vibrational energies are present.`

The concrete worry is the ordinary case: an agent optimizes to a saddle point, or the optimizer hits `steps` before `fmax`, then asks for a reaction free energy. It now gets a confident number instead of a refusal, and `raw_imaginary_mode_count` — the only remaining trace — deliberately raises no warning per 3ce56600e.

I don't think you need to resolve full mode identification to fix this. Options roughly in order of how much they ask of you:

1. **Warn (or fail) when a retained mode is implausibly soft.** A retained frequency below ~10-20 cm-1 in a molecular calculation is almost always a rigid-body residue, not a vibration. A note in `thermochemistry["warnings"]` would at least surface it.
2. **Warn whenever selection excludes a mode with `|imag|` above a physical threshold** — distinct from the blanket warning you removed in 3ce56600e, which fired on any imaginary content. ~60 cm-1i is not numerical noise.
3. **Compute the selection locally.** `Vibrations.get_energies()` comes from `np.linalg.eigh`, so the array is already ascending in `(f**2).real` and the retained set is `[i for i in range(3N - num_vibs, 3N) if np.real(all_energies[i]) > 0]`. Passing that with `vib_selection="exact"` makes indices known by construction rather than reverse-engineered, and lets you decide imaginary-mode handling explicitly instead of inheriting it.

Whichever you pick, the invariant worth holding is that a structure with a real imaginary mode must not silently produce a Gibbs energy.

Separately on the reconstruction itself: it currently depends on `_clean_vib_energies` ending with `np.real(...)`, on the sort being stable, and on the tail slice — all internal to a dependency pinned `ase>=3.29.0` with no upper bound. The key is also spelled `float(np.real(energy))` when building and `float(energy)` when looking up. A `<3.30` ceiling would make the coupling explicit.

src/chemgraph/tools/ase_core.py
```diff
@@ -247,8 +248,114 @@ def _vibrational_mode_indices(atomsdata: AtomsData, total_modes: int) -> list[in
...
+    for note in notes:
+        logger.warning(note)
+
+    return {
+        "enthalpy": enthalpy,
+        "entropy": entropy,
+        "gibbs_free_energy": gibbs_free_energy,
+        "unit": "eV",
+        "entropy_unit": "eV/K",
+        "ase_version": ase.__version__,
```
keceli, line 345: The ASE version is recorded here but never checked, and `vib_selection` does not exist before 3.29.

`pyproject.toml` now says `ase>=3.29.0` and both Dockerfiles pin `3.29.0`, but a version bump does not upgrade an already-provisioned environment. This repo's own `.venv` still has 3.25, and the remote paths are worse: `hpc_configs/loader.py:40-50` builds `worker_init` as `source <venv>/bin/activate` / `conda activate $CONDA_DEFAULT_ENV`, `aurora_parsl.py:37` falls back to `module load frameworks`, and Globus-Compute endpoints are provisioned separately. None of those are upgraded by this PR.

On any of them, every `thermo` run now returns:

```
{'status': 'failure', 'error_type': 'TypeError',
 'message': "Thermochemistry failed: IdealGasThermo.__init__() got an
             unexpected keyword argument 'vib_selection'. ..."}
```

The broad `except Exception` at line 895 turns a dependency problem into a per-molecule chemistry failure, so an agent reads it as "this molecule failed" and retries the calculation rather than surfacing "upgrade ASE". Monatomic thermo is newly affected too — on main it never constructed `IdealGasThermo` at all.

An explicit `ase.__version__` precondition before the call, raising a message that names the required version, would make this diagnosable. Worth noting `"highest"` is already ASE 3.29's default, so the explicit kwarg buys nothing except the hard incompatibility — omitting it would widen the compatible range at no cost.

src/chemgraph/tools/ase_core.py
```diff
@@ -247,8 +248,114 @@ def _vibrational_mode_indices(atomsdata: AtomsData, total_modes: int) -> list[in
...
+            type(calc_model).__name__,
+        )
+        multiplicity = 1
+
+    with warnings.catch_warnings(record=True) as caught:
+        warnings.simplefilter("always", UserWarning)
+        thermo = IdealGasThermo(
+            vib_energies=all_energies,
+            potentialenergy=potential_energy,
+            atoms=atoms,
```
keceli, line 309: Passing the live `atoms` breaks every periodic input under the new ASE floor.

ASE 3.29 added a guard to `BaseThermoChem.__init__` that 3.25 did not have:

```python
if atoms.pbc.any():
    raise ValueError("Atoms object should not have periodic boundary conditions "
                     "for BaseThermoChem.")
```

`atoms` comes straight from `ase.io.read(input_structure_file)`, which sets `pbc=True` for CIF, POSCAR, extxyz-with-Lattice and `.traj` — all formats `ASEInputSchema.input_structure_file` advertises. So `driver="thermo"` on any of them now returns `status: failure` where it previously returned numbers.

The rest of the function already treats the system as an isolated molecule — `get_symmetry_number` explicitly documents that it wants unwrapped, non-periodic coordinates — so passing a pbc-stripped `atoms.copy()` here is consistent with what the surrounding code already assumes, and avoids a failure that has nothing to do with the user's input being wrong.

src/chemgraph/tools/ase_core.py
```diff
@@ -247,8 +248,114 @@ def _vibrational_mode_indices(atomsdata: AtomsData, total_modes: int) -> list[in
...
+        )
+    if len(atoms) == 1:
+        geometry, symmetrynumber = "monatomic", 1
+    else:
+        geometry = "linear" if is_linear_molecule(final_structure) else "nonlinear"
+        symmetrynumber = get_symmetry_number(final_structure)
+
+    # IdealGasThermo expects total spin S; calculators expose 2S+1.
+    multiplicity = getattr(calc_model, "get_multiplicity", lambda: None)()
+    if multiplicity is None:
```
keceli, line 295: Narrowing `or 1` to `is None` lets `multiplicity=0` through, and it lands as a non-finite entropy.

The old `getattr(calc_model, "get_multiplicity", lambda: None)() or 1` coerced a falsy 0 to 1. Now 0 survives to `spin = (0 - 1) / 2.0 = -0.5`, and ASE computes `S_e = kB * log(2 * spin + 1)` = `log(0)` = `-inf`. The new `np.all(np.isfinite(...))` check then raises "ASE returned non-finite thermochemistry values." — a message that gives no hint the multiplicity input was the cause.

`multiplicity=0` is reachable: `TBLiteCalc.multiplicity` and `OrcaCalc.multiplicity` have no `ge=1` constraint (only NWChem and MACE do), and `FAIRChemCalc` has a deprecated-alias validator mapping a supplied `spin` key onto `multiplicity`. Since ASE's own parameter is named `spin` and 0 means singlet there, `{"calculator_type": "fairchem", "spin": 0}` is a natural thing for a model to emit — and it silently produced the correct singlet answer on main.

Either restore the falsy check, or add `ge=1` to the TBLite/ORCA fields and raise a targeted error naming multiplicity.

src/chemgraph/tools/ase_core.py
```diff
@@ -708,21 +820,53 @@ def run_ase_core(params: ASEInputSchema) -> dict:
...
+                "energy_unit": "meV",
+                "frequencies": [],
+                "frequency_unit": "cm-1",
+            }
             mol_stem = (
                 Path(input_structure_file).stem if input_structure_file else "mol"
             )
+            # Remove previous modes even when this run has no vibrations.
+            freq_file = Path(_resolve_path(f"frequencies_{mol_stem}.csv"))
+            freq_file.unlink(missing_ok=True)
```
keceli: This is the one I caused — I asked for the cleanup to be hoisted above the guard, and hoisting it this far opened a data-loss window.

Lines 843-850 now unlink `frequencies_<stem>.csv` and glob-unlink `<stem>_vib.*.traj` (and for `ir`, the spectrum PNG/CSV and peaks CSV) *before* the displacement loop runs. Previously the CSV was unlinked immediately before being rewritten, and the trajectories were removed after `write_mode` had produced replacements.

So a calculator failure partway through the 6N displacements — an SCF non-convergence with NWChem/ORCA, an OOM, a walltime kill — now returns a failure with the previous successful run's frequencies, mode animations and IR artifacts already deleted. On main those survived. The UI then shows "Frequencies file not found." for a molecule the user had good data for minutes earlier.

The single-atom case I was actually asking about doesn't need this reach. Either scope the unconditional cleanup to `len(atoms) == 1`, or keep it where it was and add a monatomic-only cleanup branch.

Related, at line 897: on a thermo failure `mode_indices` is set to `[]`, so the writer at 914 opens the CSV in `"w"` mode and writes zero rows, and the `write_mode` loop copies nothing — while the message says "Completed structure, energy, and vibration results saved to ...". The Hessian was computed and lives in `all_modes`, but the CSV and trajectories the message points at are empty.

src/chemgraph/tools/report_tools.py
```diff
@@ -616,34 +652,56 @@ def add_additional_info_to_html(html_content: str, ase_output: ASEOutputSchema)
     # Thermochemistry Values
     if ase_output.thermochemistry:
         thermo_info = []
+        if "ase_version" in ase_output.thermochemistry:
+            metadata = ase_output.thermochemistry
+            thermo_info.append(
+                f"<div>ASE {escape(str(metadata['ase_version']))}; "
+                f"mode selection: {escape(str(metadata['vib_selection']))}; "
+                f"ignore_imag_modes: {escape(str(metadata['ignore_imag_modes']))}.</div>"
+                f"<div>Imaginary modes in the complete input: {metadata['raw_imaginary_mode_count']}. "
```
keceli: These two values are interpolated unescaped, three lines after their siblings are deliberately escaped.

`ase_version`, `vib_selection` and `ignore_imag_modes` above all use `escape(str(...))`, and the warnings loop below does too — but `metadata['raw_imaginary_mode_count']` and `metadata['n_imag']` go in raw.

`ASEOutputSchema.thermochemistry` is a bare `dict` with no type validation, and `generate_html` is an LLM-callable tool that accepts an arbitrary `results_json_path`. A results file containing `"n_imag": ""` renders that markup live into the report. Given the surrounding lines already escape, this reads as an oversight rather than a decision.

src/chemgraph/tools/report_tools.py
```diff
@@ -616,34 +652,56 @@ def add_additional_info_to_html(html_content: str, ase_output: ASEOutputSchema)
     # Thermochemistry Values
     if ase_output.thermochemistry:
         thermo_info = []
+        if "ase_version" in ase_output.thermochemistry:
```
keceli: One key-presence check unlocks five direct subscripts.

This guard tests only `ase_version`, then the block indexes `metadata['vib_selection']`, `metadata['ignore_imag_modes']`, `metadata['raw_imaginary_mode_count']` and `metadata['n_imag']` with `[]` — while the very next lines defensively use `metadata.get("warnings", [])` and `.get("entropy_unit", "eV/K")`.

Any payload carrying `ase_version` without the full sibling set raises `KeyError`. That's reachable from a hand-edited file, a truncated run, or a future version that renames one of these keys. Because of the swallow described in the `entropy_unit` comment, the user gets no report and the agent believes one was written.

`.get()` with sensible fallbacks for all five would match the style already used two lines down.

src/chemgraph/tools/report_tools.py
```diff
@@ -555,18 +558,19 @@ def add_additional_info_to_html(html_content: str, ase_output: ASEOutputSchema)
...
-                includes_nonvibrational_modes and i <= trans_rot_modes
+                not ase_selected and includes_nonvibrational_modes and i <= trans_rot_modes
             )
             mode_type = (
                 "Translation/Rotation" if is_nonvibrational else "Vibrational"
             )
             row_class = (
                 "trans-rot-mode" if is_nonvibrational else "vibrational-mode"
             )
+            display_index = selected_indices[i - 1] + 1 if ase_selected else i
```
keceli: `ase_selected` is derived from `all_modes` but this line indexes `mode_indices`, which the adjacent line treats as optional.

Line 533 reads `selected_indices = ...get("mode_indices", [])` — so an absent key is legal — but the loop length comes from `frequencies`, not from `selected_indices`. The `.get(..., [])` default converts a `KeyError` into an `IndexError` rather than making the loop safe.

This is reachable from ChemGraph's own writer: on a thermo failure, `ase_core.py:897` sets `mode_indices = []` while `all_modes` is already populated, so a failed run carries the exact signature of a successful one. Removing only `mode_indices` from a real water result gives `Failed to generate HTML report: list index out of range`, and no file is written — the structure viewer, energy and spectrum are all lost over a missing diagnostic key.

Iterating `zip(selected_indices, frequencies, energies)`, or falling back to `i` when the lengths disagree, would keep the report renderable.

src/chemgraph/tools/report_tools.py
```diff
@@ -616,34 +652,56 @@ def add_additional_info_to_html(html_content: str, ase_output: ASEOutputSchema)
...
             thermo_info.append(
-                f'<div><strong>Enthalpy:</strong> <span class="energy-value" data-ev="{enthalpy_ev}">{enthalpy_ev:.6f}</span></div>'
+                f'<div><strong>Enthalpy:</strong> <span class="energy-value" data-ev="{enthalpy_ev}">{enthalpy_ev:.6f}</span> <span class="energy-unit">eV</span></div>'
             )
         if "entropy" in ase_output.thermochemistry:
-            entropy_ev = ase_output.thermochemistry['entropy']
+            entropy_unit = ase_output.thermochemistry.get("entropy_unit", "eV/K")
+            if entropy_unit != "eV/K":
+                raise ValueError(
+                    f"Unsupported entropy_unit {entropy_unit!r}; expected 'eV/K'. "
```
keceli, line 728: Aborting the whole report over one metadata string is a very blunt failure mode — and the abort is then misreported as success.

The raise happens inside `add_additional_info_to_html`, so `generate_html`'s `except Exception` at line 434 converts it to the string `"Failed to generate HTML report: ..."` and writes nothing. The enthalpy, Gibbs energy, structure viewer, convergence state and full spectrum were all renderable.

The part worth fixing regardless of what you decide about strictness: `single_agent._is_successful_report_message` classifies that string as **success**, because it checks `not content_text.startswith("error")` and the message starts with "failed". `route_after_report_tools` then returns `"done"`. So the agent terminates believing the report exists when no file was written. This PR adds three new ways to reach that path (this raise, the `KeyError` above, the `IndexError` at line 569), which is what turns a display concern into a silent workflow failure.

Two suggestions, independent of each other:

1. Degrade rather than abort — render the entropy row with its declared unit and skip only the conversion. Better still, validate at write time with a `Literal["eV/K"]` on a typed thermochemistry model, where the producer can still act on it.
2. Make `_is_successful_report_message` check for the actual sentinel (or have `generate_html` return a string starting with "Error:"), so a failed report routes to `retry` instead of `done`.

tests/test_report.py
```diff
@@ -166,6 +243,144 @@ def test_generate_html_labels_energy_by_driver(
...
+    assert not report.exists()
+    report.write_text("previous report")
+    generate_html.invoke({"results_json_path": str(source), "output_path": str(report)})
+    assert report.read_text() == "previous report"
+
+
+@pytest.mark.parametrize("script_tag", ["script", "SCRIPT", "ScRiPt"])
+def test_report_unit_conversion_javascript(tmp_path, script_tag):
+    """Execute the generated converter, with only its DOM dependencies stubbed."""
+    quickjs = pytest.importorskip("quickjs")
```
keceli, line 537: This test never runs anywhere but your machine — `quickjs` is not declared in `pyproject.toml`, any extra, `requirements/`, `environment.yml`, or any CI step.

Locally: `pytest tests/test_report.py` gives `19 passed, 3 skipped`, with `SKIPPED [3] could not import 'quickjs'`. CI installs `-e .` plus `ruff`, so it skips there too.

That matters because this is the only coverage of the new `.entropy-value` / `dataset.evK` / precision path, which is exactly what the last round asked for. A regression like reading `cell.dataset.ev` instead of `cell.dataset.evK` renders every entropy cell as `NaN` after a unit toggle and ships green.

Adding `quickjs` to a declared test extra that `tests.yml` installs would make the coverage real. Failing that, dropping `importorskip` at least makes the gap visible rather than silent.

keceli (changes requested):
Re-reviewed against `3ce5660`. **Every fix from the last round is verified**, and several are better than what I suggested:

- `temperature: float = Field(default=298.15, gt=0, allow_inf_nan=False)` plus the null-coercing before-validator, on both schemas.
- `verbose=False` on both getters with `G = H - T*S`, and the non-finite guard.
- `get_symmetry_number` returns 1 for a single atom; cleanup hoisted above the guard with `glob.escape`; guard now keys on `len(atoms) > 1` for all three drivers.
- Renderer reads `entropy_unit`; `.energy-value, .entropy-value` share the sizing rule; 6/4 precision; "Units:" label.
- Tests use the real `MaceCalc` accessor, and the complex-spectrum cases with pinned expected indices (`selected-zero`, `boundary-duplicates`, `unordered`) are genuinely discriminating — that is the coverage I was asking for.

You were also right to reject my `energies[6:]` suggestion. Deriving the reported modes from what ASE retained is the better design.

**One blocker, and it's the policy question you left open.** I've answered it in [that thread](https://github.com/argonne-lcf/ChemGraph/pull/233#discussion_r3943243659) rather than here: `vib_selection="highest"` + `ignore_imag_modes=True` makes ASE drop genuine imaginary modes during *selection*, before `_clean_vib_energies` runs, so `n_imag` stays 0 and nothing warns. Your `recorded-cu4` fixture encodes it — two modes at ~60 and ~32 cm-1 imaginary vanish silently while two 0.74 cm-1 residues are retained as vibrations, worth ~0.34 eV of spurious `T*S`. On main that input failed loudly. A saddle point now returns a confident Gibbs energy.

The rest are inline.

### Not inline (files outside this diff)

- **`mace_parsl_schema.py:39/99/139`** kept `temperature: float = 298.15` with no `gt=0` and no null-coercion, so the two front ends now disagree. `temperature=0` passes `mace_input_schema`, then `_mace_input_to_ase_input` (`parsl_tools.py:68`) raises a pydantic `ValidationError` inside `run_mace_core`, which has no try/except and is called directly from `mace_mcp_hpc.py:127` and `mace_mcp_parsl.py:99` — so the MCP tool raises instead of returning a failure dict. Conversely `temperature=null` now works on `run_ase` and fails on `run_mace_single`.
- **The prompts still treat temperature as required.** `prompt/multi_agent_prompt.py:30`, `:90`, `:171` and `prompt/single_agent_prompt.py:7` list a missing temperature as a reason to stop and ask the human, which `docs/calculators.md:39-41` now contradicts. "Gibbs free energy of water with MACE" will still trigger a clarification instead of running at the documented default.

One of these is mine: hoisting the artifact cleanup above the guard is what I asked for, and it introduced the data-loss window flagged at `ase_core.py:844`. Sorry for the round trip.

---

🤖 Findings generated with [Claude Code](https://claude.com/claude-code) (Opus 5) and posted by @keceli.

src/chemgraph/tools/ase_core.py
tdpham2: Updated in [437a5fdb4](https://github.com/argonne-lcf/ChemGraph/commit/437a5fdb44cbac10a1a43253fe23f9099f734d6c) to use ASE’s defaults: `vib_selection="highest", ignore_imag_modes=False`.

Imaginary modes remaining after selection now cause thermochemistry to fail. Selected zero-energy modes that produce non-finite H/S/G also fail. The existing failure path preserves the completed structure, potential energy, convergence state, and complete spectrum in JSON, returns `results_file`, and omits thermochemistry values.

The selection-stage concern you identified remains: with `highest`, Cu4’s imaginary modes are excluded before validation, so that fixture still succeeds with `n_imag=0` and `warnings=[]`. This commit does not establish structural stability or resolve that case. The complete spectrum remains available in JSON and HTML.

For this PR, I’m retaining ASE’s default selection without adding a ChemGraph frequency threshold. [ASE documents limitations of both `highest` and `abs_highest`](https://docs.ase-lib.org/ase/thermochemistry/thermochemistry.html). Identifying rigid-body motion and genuine internal modes needs a stronger basis than frequency ranking alone. I’m leaving this thread open for discussion of that scientific policy.

Updated the metadata, documentation, and regression expectations. Ruff and diff checks pass; 80 focused tests pass; the full suite passes with 874 passed, 21 skipped, and 2 deselected.

keceli:
Re-reviewed at head `d1f6e97` (11 commits, +1163/−122, 12 files). Per instruction, the
rigid-body / imaginary-mode *selection policy* thread (`highest` dropping imaginary modes before
`_clean_vib_energies`) is out of scope for this pass and is not re-litigated below.

Verification environment: fresh conda env, ASE 3.29.0, pymatgen, torch, langchain-core, `PYTHONPATH=src`.
`pytest tests/test_thermochemistry.py tests/test_report.py` → **81 passed, 3 skipped** (the three
skips are the QuickJS tests, see item 8).

## Verdict: Request changes

The PR does not merge (`mergeable: false`) and 7 of the 9 comments from the 2026-09-09 round are
unchanged at head. Each one was re-executed against `d1f6e97` rather than re-read; results are below.

## Resolved since last round

| Comment | Commit | Verified |
|---|---|---|
| Unescaped `raw_imaginary_mode_count` / `n_imag` in HTML | `d1f6e97` | Yes — both now go through `escape(str(...))`; test added. |
| `ignore_imag_modes=True` policy | `437a5fd` | Switched to `ignore_imag_modes=False` (ASE default). Not evaluated further here per scope note. |

## Blocking

**1. Merge conflict with `main` (both Dockerfiles).**
`#231` replaced the trailing `RUN` block in `Dockerfile` and `Dockerfile.arm` with a `pip check`
validation step; this branch rewrites the same block to pin `ase==3.29.0`. `git merge-tree` shows
two conflict hunks. Needs a rebase, and the rebased block should keep both the ASE pin *and*
`python -m pip check` / `import graph_longrange` from main.

**2. Periodic inputs now fail `thermo` unconditionally** (comment 3964045025, unchanged).
`_calculate_thermochemistry` passes the live `atoms` to `IdealGasThermo`; ASE 3.29's
`BaseThermoChem.__init__` raises on `atoms.pbc.any()`. Reproduced end-to-end through
`run_ase_core` with a water `.cif` written by ASE:

```
status: failure
message: Thermochemistry failed: Atoms object should not have periodic boundary
         conditions for BaseThermoChem.
```

`ase.io.read` sets `pbc=[True,True,True]` for `.cif`, `POSCAR`, `.extxyz` with a `Lattice`, and
`.traj` written from a cell-bearing Atoms — all formats the input schema advertises. Fix is one
line: hand ASE `atoms.copy()` with `pbc=False`. No test in the PR exercises a periodic input file.

**3. `multiplicity=0` produces an unattributed non-finite failure** (comment 3964045029, unchanged).
`TBLiteCalc(multiplicity=0)` and `OrcaCalc(multiplicity=0)` both validate and return 0 from
`get_multiplicity()`; the `is None` check lets it through, `spin=-0.5`, and ASE returns
`S = -inf`. Reproduced via `run_ase_core`:

```
status: failure
message: Thermochemistry failed: ASE returned non-finite thermochemistry values.
```

Nothing in that message points at multiplicity. Either restore the falsy coercion, or add `ge=1`
to the TBLite/ORCA fields and raise a targeted error.

**4. Stale-artifact cleanup runs before the displacement loop** (comment 3964045036, unchanged).
`frequencies_<stem>.csv` and `<stem>_vib.*.traj` are unlinked at the top of the `vib/thermo/ir`
block, before `Vibrations.run()`. Reproduced: a successful water run leaves
`frequencies_water.csv, water_vib.6.traj, water_vib.7.traj, water_vib.8.traj`; a second run whose
calculator raises after 10 evaluations returns `failure: SCF did not converge` and the surviving
set is `[]`. On `main` those files survived a mid-run failure. Scope the unconditional cleanup to
`len(atoms) == 1` (the case that motivated hoisting it) and keep the molecular cleanup where it was.

**5. Thermo failure writes an empty CSV while the message says results were saved** (same comment).
On the `.cif` failure above: `frequencies_wcif.csv` has 0 rows, 0 `.traj` files are written,
saved JSON has `mode_indices: []`, `all_modes: 9 entries`, `thermochemistry: {}` — and the
message ends "Completed structure, energy, and vibration results saved to .../output.json".
The Hessian is in `all_modes`; the CSV and trajectories the message points at are empty. Either
fall back to `_vibrational_mode_indices` for the CSV/trajectory writers on thermo failure, or say
so in the message.

## Should fix before merge

**6. `add_additional_info_to_html` aborts the whole report, and the abort is classified as success**
(comments 3964045047 / 3964045049 / 3964045051, unchanged). All three paths re-triggered on a real
water result:

| Mutation of a valid results JSON | `generate_html` return | HTML written |
|---|---|---|
| remove `thermochemistry.n_imag` | `Failed to generate HTML report: 'n_imag'` | no |
| remove `vibrational_frequencies.mode_indices` | `Failed to generate HTML report: list index out of range` | no |
| `entropy_unit: "J/(mol K)"` | `Failed to generate HTML report: Unsupported entropy_unit ...` | no |

`single_agent._is_successful_report_message` returns `not content_text.startswith("error")`, so
every one of these strings routes to `"done"`. The guard at line 655 should use `.get()` for all
five keys (matching the `.get("warnings", [])` two lines below); the loop at 569 should
`zip(selected_indices, ...)` or fall back to `i`; and either the report should degrade rather than
raise on `entropy_unit`, or `generate_html` should return a string starting with `Error:` so the
router can see it. The routing fix is independent of the strictness decision and is the part that
turns a display issue into a silent workflow failure.

**7. No ASE version precondition; `vib_selection=` kwarg is passed explicitly** (comment 3964045018,
unchanged). `ase_version` is recorded in the payload but never checked. On any environment still
at 3.25 (existing venvs, `hpc_configs/loader.py` `worker_init`, Globus Compute endpoints), every
`thermo` run — including monatomic, which never touched `IdealGasThermo` on main — returns
`TypeError: ... unexpected keyword argument 'vib_selection'` wrapped as a per-molecule chemistry
failure. Since `"highest"` and `ignore_imag_modes=False` are both 3.29 defaults, dropping the two
kwargs and adding an explicit `ase.__version__ >= 3.29` check with a message naming the required
version would make the failure diagnosable and widen the compatible range.

**8. QuickJS tests never run in CI** (comment 3964045055, unchanged). `quickjs` is not in
`pyproject.toml`, any extra, `environment.yml`, or `.github/workflows/tests.yml` (which does
`pip install -e .` / `-e ".[academy,parsl,globus_compute]"`). Locally: `SKIPPED [3] could not import
'quickjs'`. These are the only tests of the `.entropy-value` / `dataset.evK` / precision path.
Declare it in a test extra that CI installs, or drop `importorskip` so the gap is visible.

## Minor

**9. PR description is stale after `437a5fd`.** It still says `vib_selection="highest",
ignore_imag_modes=True` and "ASE's `n_imag` counts cleanup removals after selection, including
zero-energy modes". Code, schema description, and `docs/calculators.md` now say `False` and
"n_imag is zero on success". Please sync before merge since the description becomes the squash
message.

**10. HTML metadata line still describes the old policy.** `report_tools.py:663` renders
"ASE cleanup removed {n_imag} non-positive modes (imaginary or zero energy) after selection".
With `ignore_imag_modes=False`, ASE 3.29 removes nothing: an imaginary survivor raises, and a
zero-energy survivor is *retained* (verified: `vib_energies=[0.0, 0.45, 0.46]` → `n_imag=0`,
`vib_energies=[0, 0.45, 0.46]`, `S=nan`, `H=nan`) and then hits the non-finite check. For every
new result this line reads "removed 0 non-positive modes", which is true but implies a cleanup
step that no longer exists. Suggest keying the wording on `ignore_imag_modes`.

## What is right

The core fix stands: monatomic thermo through `IdealGasThermo(geometry="monatomic")` (verified
against ASE directly in the tests for multiplicity 1/2/3), `charge_spin_check=False` for radicals
(closes #141), single-atom short-circuit for all three drivers, temperature normalization with
`gt=0` / `allow_inf_nan=False`, entropy-unit contract, retained-mode indices derived from what ASE
actually kept (with the degenerate-boundary handling), the complete spectrum preserved in JSON/HTML,
and failure paths that keep the completed structure and energy. The new test file is substantive:
507 lines, controlled complex spectra checked against ASE independently, and exact CSV/trajectory
index assertions.

tdpham2:
2 should not be a problem for this PR. We only focus on gas molecules (hence the use of IdeaGasThermo). 

7 should already be addressed since we bump ASE >=3.29. User need to update their envs.

keceli (approved):

PR #237, opened 2026-09-12
Bundle portable skills for Deep Agents

tdpham2:
## Summary

Deep Agents previously required callers to provision skill sources themselves, and checkpointed catalogs could hide later updates. This PR bundles maintained skills, refreshes discovery between turns, and lets CLI users load skill collections anywhere on the host. For example, a workspace at `.` can use AtomisticSkills at `../external/AtomisticSkills/.agents/skills/` without copying files or creating symlinks.

- Ship `chemgraph` and `pbs-hpc` skills and supporting resources in wheels and source distributions, including Aurora operations guidance. Serve packaged resources through a read-only `/chemgraph-skills/` backend.
- Discover bundled, personal, project, explicit host, then explicit Python backend sources; later matching skill names win. Refresh the catalog before each turn and in delegated workers. Report required-source failures before the model call while isolating optional-source failures.
- Resolve CLI/TOML skill directories against the invocation directory, supporting absolute paths, relative paths, `~`, spaces, and existing symlinks. Mount canonical directories automatically through stable backend routes, independently of the workspace and automatic discovery setting.
- Persist canonical host directories and personal roots across supported session restoration and interactive rebuilds. Preserve legacy backend paths, fingerprints, existing routes, the default executor, and action approvals.
- Share the Deep Agent prompt across standalone and supervisor-hosted workers, preserving custom and empty overrides. Permit the Codex adapter to request all exposed ChemGraph tools and retain tool-call/result identities in conversation history. ChemGraph still executes and reviews requests; Codex-native tools remain disabled.

## CLI example

With an existing AtomisticSkills checkout beside the ChemGraph workspace:

```sh
chemgraph run --interactive --workflow deep_agent --deepagent-workspace . \
  --deepagent-skill ../external/AtomisticSkills/.agents/skills/ \
  --model argo:gpt-5.6-luna
```

Ask: “Read the chem-sorption-widom skill and identify its required tools, environments, and script paths before running anything.”

The collection must contain immediate `<skill-name>/SKILL.md` directories. Registration loads instructions and resources; it does not install environments or attach MCP tools. Local shell path mappings are provided in the prompt; remote executors require their own file visibility.

## Compatibility

- CLI `--deepagent-skill` and TOML `deepagent_skills` now mean host directories. Replace older CLI `/workspace/...` examples with actual host paths. CLI flags replace the TOML list.
- Python `skills` / `deepagent_skills` retain their backend-relative meaning. New `skill_dirs` / `deepagent_skill_dirs` explicitly mount host collections, including with state/store/remote defaults.
- Existing empty filesystem collections remain valid. Automatic personal/project discovery can be disabled independently of explicit sources.

## Validation

- `ruff check .` and `git diff --check` passed.
- Full Python 3.12 suite: **1,229 passed, 21 skipped, 2 deselected**, using `pytest tests/ -k "not tblite"` with the pinned dependencies. Local MCP tests ran with localhost access; live-service test flags remained disabled.
- Hermetic coverage includes path forms and errors, source precedence, sync/async reads of external Widom instructions and helpers, delegated workers, prompt path mappings, backend preservation, session restoration, interactive model/workflow changes, and legacy compatibility. Existing Codex tests cover tool history and approved/rejected actions.
- Read-only check against the actual external AtomisticSkills checkout: **131 skills, zero loading errors**, with successful reads of `chem-sorption-widom/SKILL.md` and `scripts/run_widom.py`. No simulation or live model call was made in this check.

## Checklist

- [x] Targets `main` through the existing PR branch
- [x] Tests cover changed behavior
- [x] Lint and core suite pass
- [x] CLI, TOML, Python, and skills documentation updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ Useful run options include:
 | `-r`, `--report` | Allow generation of an HTML report |
 | `--human-supervised` | Allow supported workflows to pause for input |
 | `--deepagent-workspace PATH` | Set the workspace for `deep_agent` or the optional main-agent worker |
+| `--no-deepagent-discover-skills` | Disable personal/project skill discovery; retain bundled and explicit skills |
 | `--deepagent-skill PATH` | Add an explicit backend-relative Agent Skills directory; repeat to layer sources |
 | `--output-file` | Save the CLI response to a file |
 | `-v` / `-vv` | Enable INFO / DEBUG diagnostics |
@@ -165,9 +166,12 @@ chemgraph run --interactive --workflow main_agent --deepagent \
   --deepagent-skill /workspace/.agents/skills/
 ```
 
-Skill directories are never discovered automatically. Each source contains
-one directory per skill with a `SKILL.md`; when names collide, the source from
-the later `--deepagent-skill` occurrence wins.
+ChemGraph bundles `chemgraph` and `pbs-hpc` skills and automatically discovers
+`~/.chemgraph/skills/` and the workspace's `.agents/skills/`. Additional
+`--deepagent-skill` sources override earlier matching names. Use
+`--no-deepagent-discover-skills` to disable personal/project discovery. Skills
+use one directory per `SKILL.md`; their catalog refreshes each new turn. See
+[skills and filesystem backends](docs/skills.md).
 
 Headless workspace mutation is disabled unless the command includes an
 explicit workspace and `--deepagent-dangerously-skip-approvals`. Use that mode
@@ -254,7 +258,7 @@ stdio client configuration and the experimental HPC servers.
 | --- | --- | --- |
 | `single_agent` | General molecule lookup, ASE calculations, and reports | Default and recommended first workflow |
 | `main_agent` | Long-lived supervisor with delegated chemistry work | Interactive mode; use `MainAgentSession` in Python |
-| `deep_agent` | Repository exploration, coding, and workspace tasks (`deepagent` is an alias) | Interactive approvals by default; broad local shell access |
+| `deep_agent` | Workspace tasks and attached chemistry tools (`deepagent` is an alias) | Interactive approvals by default; broad local shell access |
 | `multi_agent` | Planner/executor decomposition and parallel subtasks | More model calls and orchestration overhead |
 | `python_relp` | LLM-directed Python and arithmetic (`python_repl` is an alias) | Executes Python in the ChemGraph process; use only with trusted prompts |
 | `molecular_docking` | Ligand/receptor docking with AutoDock Vina | `docking` extra plus Vina from conda-forge |
--- a/config.toml
+++ b/config.toml
@@ -11,7 +11,9 @@ human_supervised = false
 # To call it directly, set workflow = "deep_agent" instead.
 enable_deepagent = false
 # deepagent_workspace = "."
-# Ordered backend-relative Agent Skills directories (later entries win).
+# Automatically discover ~/.chemgraph/skills and <workspace>/.agents/skills.
+deepagent_discover_skills = true
+# Additional backend-relative skill directories (later entries win).
 # deepagent_skills = ["/workspace/.agents/skills/"]
 # Durable main-agent checkpoint database.
 # checkpoint_db = "~/.chemgraph/checkpoints.db"
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -110,19 +110,17 @@ The selected directory is mounted for file tools at `/workspace`. Thus,
 absolute host-path mapping supplied to the model. Existing files under a
 previously created `test/workspace/` directory are not migrated.
 
-Skills are opt-in. Repeat `--deepagent-skill PATH` to provide ordered,
-backend-relative source directories; the later source wins when two sources
-contain the same skill name. For the virtual CLI workspace, a conventional
-project source can be selected explicitly as
-`/workspace/.agents/skills/`. ChemGraph does not scan project or user
-directories automatically.
+Bundled `chemgraph` and `pbs-hpc` skills load automatically, followed by
+`~/.chemgraph/skills/` and the workspace's `.agents/skills/` when present.
+Repeat `--deepagent-skill PATH` to add backend-relative sources; later sources
+win for matching skill names. `--no-deepagent-discover-skills` disables personal
+and project discovery while retaining bundled and explicit sources.
 
-Each source must contain one directory per skill, and each skill directory
-must contain a `SKILL.md` with `name` and `description` YAML frontmatter.
-Metadata is cached in the checkpoint for the life of the thread, so restart or
-reinitialize the workflow after changing the available skill set. Reading a
-skill does not require an action review, while executing a bundled script or
-mutating its files continues to use the normal Deep Agent approval policy.
+Each source contains a directory per skill with a `SKILL.md` and `name` and
+`description` YAML frontmatter. Metadata refreshes at each new turn, including
+on reconstructed graphs. Bundled resources are read-only; templates/helpers
+must be copied into the execution filesystem before use by the shell. Normal
+file mutation and execution approvals apply. See [skills](skills.md).
 
 Deep Agent run logs use the normal ChemGraph locations rather than the selected
 workspace. The default is `cg_logs/session_*` for state JSON plus the configured
--- a/docs/configuration_with_toml.md
+++ b/docs/configuration_with_toml.md
@@ -111,6 +111,7 @@ definitions. See [MCP servers](mcp_servers.md).
 workflow = "main_agent"
 checkpoint_db = "~/.chemgraph/checkpoints.db"
 enable_deepagent = false
+deepagent_discover_skills = true
 # deepagent_skills = ["/workspace/.agents/skills/"]
 ```
 
@@ -124,7 +125,11 @@ The headless-only `--deepagent-dangerously-skip-approvals` switch is
 intentionally not configurable through TOML. It must be typed explicitly for
 each run together with `--deepagent-workspace`.
 
-`deepagent_skills` is an ordered list of backend-relative source directories.
+`deepagent_discover_skills` defaults to true and controls personal/project
+skill discovery for local workspaces. Bundled skills are always available.
+The matching CLI boolean flag overrides TOML. See [skills](skills.md).
+
+`deepagent_skills` is an ordered list of additional backend-relative sources.
 It applies to a direct `deep_agent` or to an enabled `main_agent` worker. Later
 sources override earlier sources with the same skill name. A repeated
 `--deepagent-skill` CLI option replaces the TOML list for that run; explicitly
--- a/docs/python_api.md
+++ b/docs/python_api.md
@@ -112,14 +112,22 @@ A virtual `LocalShellBackend` is exposed to the agent at `/workspace`, so file
 tool path `/workspace/src/example.py` maps directly to
 `/path/to/checkout/src/example.py`. The generated Deep Agents system context
 also supplies that host path for shell commands. Custom backends, existing
-composite backends, and non-virtual local backends are passed through unchanged.
+composite routes, and non-virtual local execution are preserved when adding
+the bundled skill route.
 
-`deepagent_skills` contains ordered POSIX paths interpreted by that same
-backend. ChemGraph passes them to Deep Agents without scanning any implicit
-user or project locations. Each source contains skill directories with a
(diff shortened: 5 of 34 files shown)


keceli (changes requested):
Base `main @ 6599963f` (post-#219). 14 commits, +3053/−108, 39 files. `mergeable_state: clean`. First
maintainer review; one Codex bot comment (P2) on `skills/runtime.py:212`.

Verification: v0.7.0-pinned env (`deepagents 0.7.5`, `langgraph 1.2.11`), `PYTHONPATH=src`.
- 12 affected test modules (`test_skills`, `test_skill_backend`, `test_skill_distribution`, `test_external_skills`,
  `test_deep_agent*`, `test_main_agent*`, `test_codex_model`, `test_llm_agent`, `test_agent_registry`): **285 passed**.
- Full suite `-k "not tblite"`: 1155 passed, 30 skipped, 21 failed — the same 21 environmental failures seen on
  #219 (MACE download, `ProcessPoolExecutor`, `~/.chemgraph`, subprocess without install); none in PR-touched files.
- `ruff check .` clean. Wheel built from the head contains all 11 `chemgraph/skills/**` files (two `SKILL.md`,
  five references, one asset, three modules).

## Verdict: Request changes (one small regression), plus one design change I'd like before merge

### 1. [P2 — regression of #219] A stale saved skill directory blocks non-Deep-Agent interactive runs

`cli/main.py` now calls `resolve_skill_dirs(deepagent_skills)` whenever the value is non-`None`, and
`commands.interactive_mode` resolves `deepagent_skill_dirs` before it looks at the workflow. In #219 we made
saved `deepagent_skills` inert for other workflows ("retained for later interactive switches"); this PR validates
them eagerly, and `resolve_skill_dirs` is strict (`resolve(strict=True)` + `os.scandir`).

Reproduced with `general.workflow = "single_agent"`, `enable_deepagent = false`,
`deepagent_skills = ["<dir that no longer exists>"]`:

```
chemgraph run --config config.toml -q test                 -> dispatched single_agent, skill_dirs=None   (ok)
chemgraph run --config config.toml -q test --interactive   -> "Invalid Deep Agent skills: Cannot access skill
                                                              directory '.../moved-away' ..."  exit 2
```

`test_saved_deepagent_settings_do_not_block_other_workflows` passes only because its saved path is `tmp_path`,
which exists. Suggested fix: resolve host directories only when `uses_deepagent` (one-shot) or at the moment the
interactive session switches to a Deep Agent workflow; for retained-but-unused values, warn and drop. Add the
missing-directory case to the test.

### 2. [P2 — design] Personal, project and external skill mounts are writable through the file tools

`BundledSkillsBackend` is read-only (verified: `write` → "Bundled skills are read-only.", `../` → `invalid_path`).
The other three sources are plain `FilesystemBackend(virtual_mode=True)` routes, so `write_file`/`edit_file`
succeed on them. Reproduced on an explicit `--deepagent-skill` collection:

```
comp.write("/chemgraph-external-skills/<digest>/widom/SKILL.md", "... ALWAYS run rm -rf first")  -> success
host file now ends with: ALWAYS run rm -rf first
```

Skill files are instructions the agent re-reads every turn ("refresh the catalog before each turn"), so this is
the agent editing its own future instructions on the host — and, with `--deepagent-dangerously-skip-approvals`, with
no approval at all. The unconfined shell can do the same, so this is not a new capability, but the file-tool path is
the one the model is told to use and the one that survives into a shared AtomisticSkills checkout. A ten-line
read-only adapter (reuse the `BundledSkillsBackend` error strings) wrapping the three `FilesystemBackend` routes
would make skill sources uniformly read-only; the docs' "Existing write/execute approvals remain in force"
sentence (`docs/skills.md:164-166`) could then say "skill routes are read-only".

### 3. [P3] Codex's comment on `runtime.py:212` — correct in shape, not reachable with deepagents 0.7.5

I tested the scenario: a project or explicit directory holding one valid skill and one malformed `SKILL.md`
(name mismatch, invalid YAML, missing description). Upstream `SkillsMiddleware` drops the malformed skill
*silently* and never emits `skills_load_errors` for it; errors are source-level only (unreadable directory), and
in that case `skills_metadata` for the source is already empty. So the early `return` at line 212 discards
nothing today. The one-line change Codex suggests (merge metadata, then record errors) is still worth taking —
it is future-proof and costs nothing — but it is not a defect at the current pin.

Related observation for the docs: because upstream is silent about malformed skills, a typo in a
`SKILL.md` frontmatter simply makes the skill vanish from the catalog. `docs/skills.md` could say so and point
at `chemgraph` … whatever listing command exists (or suggest `ls /chemgraph-external-skills/…` from the agent).

### 4. [P3] Topology fingerprint now embeds the personal skill root (an absolute `$HOME` path)

`topology_payload["deepagent_user_skills_dir"] = self.deepagent_user_skills_dir` for every local-workspace Deep
Agent session. Verified: two otherwise identical configurations under different `HOME` values produce different
fingerprints. Restoration is fine — the stored record carries the frozen path, and the legacy default
(`deepagent_discover_skills=False`) reproduces the pre-PR fingerprint exactly: I built `ChemGraph(workflow_type=
"main_agent", enable_deepagent=True)` from the base tree `6599963f` and from this head with `discover_skills=False`,
and the fingerprints are byte-identical for both the state backend (`762ee493…`) and a local-workspace
`LocalShellBackend` (`66bf39ae…`) — so this is a portability note, not a break: a session DB moved to another account or
container restores against the *old* account's personal skills path (intended, per the schema comment) but
cannot be pointed at the new one without an incompatible-topology error. If that's the intended contract, a
sentence in `docs/skills.md` under sessions would help.

### 5. [P3] Private-API coupling to deepagents

`skills/backend.py` imports `deepagents.middleware.skills._parse_skill_metadata` and four helpers from
`deepagents.backends.utils`; `runtime.py` uses `CompositeBackend._get_backend_and_key`. All fine at the exact
`deepagents==0.7.5` pin, and the test modules import them, so a bump will fail loudly at collection rather than
at run time. Worth a comment near the imports so the next person bumping the pin knows why the tests broke.

## Verified as described

- Discovery order bundled → personal → project → explicit host → explicit Python-backend, later names win
  (`test_skills.py`; reproduced: a project `good-project` and personal `good-personal` both appear alongside
  `chemgraph`/`pbs-hpc`).
- `resolve_skill_dirs`: absolute/relative/`~`/spaces/symlink handling with strict resolution; canonical digest
  routes independent of the workspace. Explicit dirs on a `StateBackend` workspace mount and load
  (`['chemgraph','pbs-hpc','widom']`).
- Per-turn refresh in `ChemGraphSkillsMiddleware.before_agent`; approvals and pending interrupts unaffected
  (`test_external_skills.py`, `test_deep_agent_review.py`).
- Legacy checkpoint restore for `main_agent` (`test_file_state_addition_restores_legacy_checkpoint`) and the
  `deepagent_discover_skills` schema default preserving old fingerprints.
- Codex adapter: tool requests now serialize `tool_calls`/`tool_call_id`/`status`, and the instruction text
  makes Codex request ChemGraph tools rather than its own; execution, approvals and backends remain ChemGraph's.
- Packaging: `[tool.setuptools.package-data] "chemgraph.skills" = [...]` — wheel verified above;
  `test_skill_distribution.py` covers sdist/wheel inclusion.
- Bundled skill content (`chemgraph`, `pbs-hpc` with Aurora/Polaris/Crux references and a PBS template) reads
  correctly and matches the current CLI flags.

tdpham2:
Thanks for the detailed review. I pushed three follow-up commits addressing the findings:

1. Stale skill directories: fixed in c1cff8670. Interactive sessions retain saved paths without checking filesystem access until Deep Agent is selected. Failed activation leaves the current workflow usable, and restoring the directory allows retry. Tests cover workflow/model switches, invocation-relative paths, and canonical-path retention.

2. Writable skill routes: retained intentionally. I’d like to preserve editing host skill collections through the existing approval-controlled file tools. The docs now explicitly describe direct host-file changes, their effect on future turns, and the implications of disabling approvals. This retains the concern you identified about modifying future instructions. Would you be comfortable keeping this behavior?

3. Partial skill loading: fixed in 5002351b5. I confirmed that malformed siblings preserve valid skills at deepagents==0.7.5. However, a custom backend returning both partial results and a source error does reproduce the loss. The fix preserves valid metadata while recording optional errors and handles error updates without metadata. Sync/async tests cover this.

4. Personal-root portability: documented in 1664d6839. Saved roots and fingerprint semantics remain unchanged.

5. Private upstream APIs: annotated. Added maintenance comments identifying the pinned dependency coupling and tests to check during upgrades.

keceli (approved):
Three commits since `b7caf20c` (`c1cff867`, `5002351b`, `1664d683`); +223/−17 across 8 files; still
`mergeable_state: clean`. Same v0.7.0-pinned env as round 1.

- 12 affected test modules: **304 passed** (285 last round; +19 new tests).
- Full suite `-k "not tblite"`: 1174 passed, 30 skipped, 21 failed — the identical environmental set as before
  (`test_mace`, `test_tools`, `test_execution`, `test_agent_logging`, `test_mcp`; none touched by the PR).
- `ruff check .` clean.

## Verdict: Approve

## Round-1 items

**1. Stale saved skill directory blocked non-Deep-Agent interactive runs — fixed (`c1cff867`), re-executed.**
Interactive runs now anchor saved paths (`_anchor_skill_dirs`: `expanduser().absolute()`, no filesystem access)
and defer the strict `resolve_skill_dirs` check to Deep Agent activation inside `initialize_agent`. With
`deepagent_skills = ["<dir that no longer exists>"]` in TOML:

| workflow | `-q` (one-shot) | `-q --interactive` |
|---|---|---|
| `single_agent`, deepagent off | dispatched, `skill_dirs=None` | **dispatched** (was exit 2), anchored path retained |
| `main_agent`, `enable_deepagent=true` | exit 2 with path-specific error (correct: Deep Agent is active) | dispatched; validation happens on activation |
| `deep_agent` (+ skip-approvals for headless) | exit 2 with path-specific error (correct) | dispatched; validation on activation |

Test `test_saved_deepagent_settings_do_not_block_other_workflows` is now parametrized over `missing=[False,
True]`, and `test_main_agent_cli.py:981` covers `interactive_mode(deepagent_skill_dirs=[<missing>])`. The
anchoring keeps invocation-relative meaning across later `cd`s in the session, which is the right fix for the
"resolve against the invocation directory" contract in the PR description.

**2. Writable personal/project/external skill routes — retained by the author, with documentation
(`1664d683`).** The author wants to keep editing host skill collections through the approval-controlled file
tools, and `docs/skills.md:168-173` now says explicitly that these routes write to the original host collection
(including shared checkouts), that approvals apply, that disabling approvals also disables review of these edits,
and that updated instructions affect future turns. That is an honest statement of the trade-off. My
recommendation: accept for this PR — the capability is approval-gated and the unconfined shell already has it —
and carry the read-only-by-default question (or a `--deepagent-skills-readonly` switch) into the skills RFC,
where it belongs alongside collection pinning and `allowed-tools`.

**3. Partial optional source dropping valid skills — fixed (`5002351b`).** `_merge_update` now records the
source errors and still merges `update.get("skills_metadata", [])`. The author confirmed my round-1 observation
(unreachable with a stock `FilesystemBackend` at deepagents 0.7.5) and added the case that *is* reachable — a
backend returning entries plus an `LsResult.error` — as `test_partial_source_results_preserve_valid_overrides`
(sync and async): the valid project override still wins, `pbs-hpc` remains, one error is recorded and logged.
Docs now state that malformed `SKILL.md` files are skipped with log warnings and never populate
`skills_load_errors`, with a pointer to `read_file` on the collection route for diagnosis.

**4. Personal root in the topology fingerprint — documented (`1664d683`).** `docs/skills.md:194-198` states
the contract: a moved session DB retains the original account's path; changing the root while restoring raises an
incompatible-topology error; use a new session for a new root. Semantics unchanged from round 1 (base-vs-head
fingerprints verified identical under the legacy default there).

**5. Private deepagents API coupling — annotated.** Comments at `skills/backend.py:17-18` and
`skills/runtime.py:186` name the pin and the tests to re-check on upgrade.

## Notes

- `interactive_mode` re-reads `agent.deepagent_skill_dirs` after Deep Agent initialization and after a workflow
  switch, so the session carries the canonical (resolved) paths once validated and the anchored paths before —
  consistent with "supported saved sessions retain canonical host paths".
- The broadened `except (TypeError, ValueError, RuntimeError, OSError)` around anchoring is wider than needed
  (`absolute()`/`expanduser()` can only raise `RuntimeError` for an unresolvable `~`), but harmless.
- Nothing else in the tree changed; round-1 "verified as described" items stand.

PR #246, opened 2026-09-18
Improve Deep Agent reviews and raise default graph limit to 200

### LINKED ISSUE #245: [Dev] Make Deepagent UX more user-friendly
A few existing issues with UX:
- Default mode for deepagent command is reject. Users need to type "approve". Update this to approve as default.
- Add additional option. Right now the return only takes approve/reject. If user wants to type something, need to interrupt manually via Ctrl+C. Need a better mechanism.

tdpham2:
## Summary

Deep Agent reviews now approve on Enter when the action's policy permits it. Typed feedback skips the action and returns instructions to the agent for a revised proposal; each action in a batch receives a separate decision.

Reviews display multiline commands, file content, and replacement diffs with terminal controls escaped across all action-derived text. Large previews retain their head and tail within 40 source lines and 8,000 characters, report omissions, and offer `v` / `view` to inspect the complete sanitized arguments in a pager before returning to the same decision. The prompt repeats the tool and file path, replacement previews identify `replace_all`, and supported TTYs discard queued input before decisions while redirected input remains unchanged.

The default graph recursion limit is 200 across CLI, Python/session APIs, Deep Agent construction, evaluation, and UI configuration. Explicit CLI limits override configuration for both flag spellings, stored session limits remain intact, and evaluation rejects limits below 1. Documentation explains that graph steps include middleware and tool execution, are not a token budget, and can increase evaluation runtime and cost. The UI accepts limits above its previous maximum of 100.

## Related issues

Closes #245.

## Type of change

- [x] Bug fix
- [ ] New feature
- [x] Docs
- [ ] Chore / refactor / CI

## How was this tested?

- `ruff check .`
- `git diff --check`
- `pytest tests/ -k "not tblite"`: **1545 passed, 21 skipped, 2 deselected**.
- Focused formatting, review CLI, and recursion-limit regressions: **164 passed**.
- Real POSIX pseudo-terminal coverage for queued Enter presses before and between reviews; mocked Windows/unsupported-terminal coverage.
- Coverage for escaped controls, bounded previews and full-view decisions, literal markup, unchanged action payloads, replacement scope, restricted approval policies, feedback, batches, and cancellation.
- CLI/configuration precedence, evaluation defaults and validation, stored-session/Python defaults, and Streamlit rendering/saving of custom limits.

Tests used the existing test environment with `PYTHONNOUSERSITE=1`, `PYTHONPATH` pointing to this checkout, and isolated session storage. Live LLM and Globus endpoint tests were not enabled.

## Checklist

- [x] Targets `main`
- [x] Review UI fixes and recursion-limit corrections are in separate focused follow-up commits
- [x] `ruff check .` passes
- [x] `pytest tests/ -k "not tblite"` passes
- [x] Added/updated regression tests
- [x] Updated user-facing documentation


keceli (changes requested):
Thanks for this. The recursion-limit change is applied consistently (no leftover 20/50 defaults in `src/`, `config.toml`, or docs), resumed sessions still use `stored_graph_config.recursion_limit`, and the decision loop enforces the allowed choices correctly. One security issue in the new preview should be fixed before merge; the rest are smaller.

### 1. Terminal escape sequences reach the terminal unescaped (should fix)

`src/chemgraph/cli/formatting.py`, `execute` and `write_file` branches of `format_action_review`.

The old panel printed `repr(args)`, which escaped control characters. The new preview passes `command` and `content` raw into `Syntax(...)` / `Text(...)`. Rich strips only BEL, BS, VT, FF and CR; it does not strip ESC.

Reproduction: a command of `rm -rf ~/data\x1b[2K\x1b[1A\recho harmless` renders with `\x1b[2K\x1b[1A` verbatim in the output. On a real terminal that erases the dangerous line and leaves only `echo harmless` visible. Same for `write_file` content; the `edit_file` diff also goes through `Syntax`. The JSON-arguments path is unaffected because `json.dumps` escapes ESC.

Since the model proposes these strings (and may have been steered by file or web content it read), the reviewer can be shown something other than what will run.

Suggested fix: neutralise C0/C1 control characters other than `\n` and `\t` before rendering, e.g.

```python
re.sub(r"[\x00-\x08\x0b-\x1f\x7f-\x9f]", lambda m: f"\\x{ord(m.group()):02x}", s)
```

and add an escape-laden case to `test_action_preview_is_literal_and_does_not_read_files`.

### 2. Enter-to-approve combined with buffered input (worth discussing)

`src/chemgraph/cli/commands.py`, the `Prompt.ask` in `_prompt_for_interrupt`.

Approve is now the default, so a newline typed while the agent was working, or a double-tap on the previous action in a batch, approves the next shell command before it has been read. I understand the default is what #245 asked for; I'd still drain pending stdin before each prompt when it is a TTY, e.g. `termios.tcflush(sys.stdin, termios.TCIFLUSH)`.

### 3. Preview has no size limit (minor)

A `write_file` with 5000 lines of content renders ~5000 terminal lines and pushes the tool name and path out of view. Consider truncating with a "… N more lines" note, or showing head and tail.

### 4. `edit_file` preview can mislead when `replace_all` is true (minor)

The diff shows one before/after snippet while the edit may touch many places. `replace_all` is in the JSON above the diff, but a note in the preview itself ("applies to all occurrences") would be clearer.

### Nits

- `main.py`: `"--recursion-limit" not in sys.argv` (pre-existing) does not recognise `--recursion-limit=33`, so a config value silently overrides the flag. The new precedence tests only cover the space-separated form.
- `eval/cli.py`: `args.recursion_limit or 200` treats an explicit `0` as unset (pre-existing).
- Raising the eval default from 50 to 200 means a looping query costs ~4x more before it stops; worth a sentence in `docs/evaluation.md`.

keceli (approved):

PR #218, opened 2026-08-25
Cache Anthropic prompts and capture token usage

tdpham2:
## Summary

- Add an Anthropic-native client subclass that places an ephemeral cache breakpoint on stable system prompts, covering Argo Claude and direct Anthropic routes.
- Normalize provider-reported input, output, total, cache-creation, and cache-read token counts into the existing `llm_call_finished` output.
- Preserve raw provider usage and keep extraction failures non-fatal.
- Keep the change output-only; this PR does not modify the dashboard UI.

## Related issues

None.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Docs
- [ ] Chore / refactor / CI

## How was this tested?

- `ruff check .` from a clean detached checkout
- `pytest tests/test_anthropic_native.py tests/test_llm_agent.py tests/test_loader_pr2.py -q`: 43 passed
- The broader core suite was attempted locally; remaining failures are unrelated environment/setup issues involving optional MACE calculators and an MCP subprocess editable import.
- No live Argo request was run because it requires credentials.

## Checklist

- [x] Branched off the latest `main` and targets `main`
- [x] PR is focused on a single logical change
- [x] `ruff check .` passes
- [x] `pytest tests/ -k "not tblite"` passes (locally blocked by unrelated optional/environment setup)
- [x] Added/updated tests for the change
- [x] Documentation is not required for this internal output addition

PR #247, opened 2026-09-18
Add token usage accounting and CLI session totals

tdpham2:
## Summary

ChemGraph now retains provider-reported usage across model calls, delegated workers, approval continuations, and retries. The CLI displays input/output/total tokens after each executed user turn and cumulative recorded session usage when interactive mode exits. Both displays use local counters: no extra model calls or conversation messages are created.

- Persist per-call usage immediately in the existing session database, with stable call/turn IDs, model and worker attribution, duration, and raw usage. Retain in-memory counts when memory is disabled or a write fails. Missing usage stays unknown or partial; cached input and reasoning output remain subsets of the totals.
- Preserve available counts on errors, cancellation, and durable main-agent restoration. Count executed retries and deduplicate repeated callbacks. Rejected queries, retries, and resumes leave the previous turn's usage and stored status unchanged.
- Expose persistent `history_unaccounted` coverage for legacy sessions. Session totals stay partial after new calls are recorded; entirely unaccounted history displays unavailable. Migration conservatively flags pre-marker transcripts even when some usage rows exist, without estimating missing calls or tokens.
- Send non-interactive usage to stderr, including failures and cancellation. Interactive totals retain copied summaries across model/workflow switches without keeping replaced agents alive, and revisiting a session replaces its cumulative summary rather than adding it again.
- Close SQLite connections after commit or rollback while preserving persistence of every usage snapshot. Callback offloading, batching, and long-session aggregation optimization remain separate work.
- Capture cumulative Codex SDK usage notifications, including available usage before SDK or response-parsing errors, instead of using only the last internal request. SDK threads remain ephemeral per model invocation; interrupted turns intentionally raise. Contract tests exercise the pinned SDK's actual notification models.
- Honor explicit per-call recursion limits without mutating the caller's configuration, including approval resumes. Stop duplicate log propagation while preserving child file handlers, and document the logging boundary for applications embedding the CLI.
- Document the APIs, CLI display, and remaining work in `docs/deepagent_overhead_followups.md`. Token budgets, prompt optimizations, and Codex lifecycle changes remain future work.

## Dependency

This PR targets `main` and depends on #246. **Merge #246 first, then merge this PR into `main`.**

The branch includes #246 at `ea4020673`, followed by the logging, recursion, token-accounting, and CLI changes, plus focused follow-up commits for connection cleanup, accounting correctness, and SDK contracts/documentation. Keep the dependency changes until #246 lands; synchronize with `main` afterward if needed before merging.

Related to #245. Anthropic caching in #218 and Codex structured-argument recovery in #240 remain separate work.

## Type of change

- [x] Bug fix
- [x] New feature
- [x] Docs
- [ ] Chore / refactor / CI

## How was this tested?

- `ruff check .`
- `git diff --check`
- `pytest tests/ -k "not tblite" -q`: **1570 passed, 21 skipped, 2 deselected**.
- Hermetic regression coverage for normalization, unknown/partial usage, concurrent workers, persistence/deduplication, failures, cancellation, approval resumes, restoration, retry accounting, configuration ownership, and duplicate logging.
- Added coverage for rejected operations preserving previous turn status, legacy-history migration and repeated restoration, persistence failures, collectible replaced owners, session revisits, stderr routing, and connection closure/commit/rollback/setup failures.
- Sync and async Codex contract tests use actual `openai-codex==0.144.4` notification models; fake streams cover cumulative totals, final-answer selection, parser/SDK failures, and interrupted turns.
- Tests use the existing environment with `PYTHONNOUSERSITE=1`, this checkout on `PYTHONPATH`, and an isolated session database. No live model calls or Globus endpoint tests were enabled.

For a manual check, run `chemgraph run --interactive --model codex:gpt-5.6-sol --workflow deep_agent` from an environment using this branch. Send two queries, then `/quit`: each response should show `Tokens:` and exit should show `Session tokens:`. Existing Codex authentication is required for that manual check.

## Checklist

- [x] Targets `main`; depends on #246 merging first
- [x] Changes are separated into focused commits
- [x] Lint and required core test suite pass
- [x] Added regression coverage and updated user-facing documentation


keceli (commented):
I reviewed the four commits on top of #246 (`ea40206..e940665`); my comments on the #246 commit are on that PR. I read the code and checked the Codex adapter's key names against the SDK's pydantic models, but did not run the test suite.

The accounting design is sound: one collector per user turn, records keyed by LangChain `run_id` so repeated callbacks update rather than add, unknown counts stay `None` instead of becoming zero, the Codex adapter's cumulative snapshot is flagged so `on_llm_end` does not bill it twice, and storage failures degrade to in-memory counts. Nothing here blocks merge; the items below are in priority order.

### 1. A query that fails early reports the previous turn's usage and rewrites its status

`src/chemgraph/cli/commands.py`, `_report_failed_usage`; `src/chemgraph/agent/llm_agent.py`, `run`; `src/chemgraph/agent/main_session.py`, `run`.

The decorator reads `owner._usage` in `finally`, calls `finish("failed")` on it, and prints it when the result is `None`. But `owner._usage` is only replaced part-way into the call:

- `ChemGraph.run` assigns the new collector after `_validate_config(config)` and `_ensure_session(query)`. If either raises (for example the new `recursion_limit must be a positive integer` check), `_usage` is still the previous query's collector.
- `MainAgentSession.run` raises its `_failed` / `_pending` / empty-message guards before `_start_usage()`.

In those cases the CLI prints the previous turn's `Tokens:` line under the new failure, and `update_usage_turn` overwrites the previous turn's stored status (`completed` or `interrupted`) with `failed`. Suggest having the decorator note the collector identity before the call and only finish/print when it changed, or resetting `_usage = None` at the top of `run`.

Related: several `return None` paths in `run_query` are user cancellations of an approval prompt rather than failures, and they are recorded as `failed` too.

### 2. `_run_codex_turn` reimplements the SDK's `thread.run` collector

`src/chemgraph/models/codex.py`.

I checked the alias keys (`turnId`, `tokenUsage`, `item.type == "agentMessage"`, `phase == "final_answer"`) against the SDK models and they match with `model_dump(mode="json", by_alias=True)`, and the final-answer selection mirrors `_final_assistant_response_from_items`. Two things to be aware of:

- It is stricter than the SDK: the SDK raises only for `TurnStatus.failed`, this raises for any status other than `completed` (so `interrupted` now raises). That is probably what you want, but it is a behaviour change worth a line in the docstring.
- It now depends on notification method strings and payload shapes rather than the `thread.run` contract. The `openai-codex==0.144.4` pin protects this; a comment next to the pin saying the adapter parses raw notifications would stop someone bumping it casually. I could only install 0.154.0 to check against, not the pinned version.

### 3. Synchronous SQLite work inside inline callbacks

`UsageCollector` has `run_inline = True` and every `on_chat_model_start`, `on_llm_end` and Codex `thread/tokenUsage/updated` event calls `SessionStore._connect()`, which opens a new connection, sets two PRAGMAs and does three `chmod` calls, then writes, all while holding the collector lock and on the event loop thread for async runs. `with self._connect() as conn` commits but does not close, so each of these also leaves a connection for the GC (pre-existing pattern, but this PR raises the call rate a lot). For Codex, consider persisting only the first and the `complete=True` snapshot rather than every update.

### 4. Session totals keep every replaced agent alive

`_interactive_usage` stores the owner objects so totals can be computed at exit. After `/model` or `/workflow` switches, the old `ChemGraph` / `MainAgentSession` (model client, graph, tools) stays referenced until the CLI exits. Snapshotting `owner.session_usage` when the owner is replaced would avoid that.

### 5. `chemgraph run` (non-interactive) now prints a `Tokens:` line to stdout

`src/chemgraph/cli/main.py`, `_handle_run`. `console` is stdout, so scripts that capture the printed result get an extra line. `--output-file` is unaffected. Consider stderr for this line in non-interactive mode, or a `--no-usage` flag.

### Nits

- `root.propagate = False` in `configure_logging` means applications that embed the CLI entry point and attach handlers to the root logger stop receiving `chemgraph.*` records. It is only called from `cli/main.py`, so library users are unaffected; worth one sentence in the docs.
- `ChemGraph._usage_turns` and `MainAgentSession._usage_turns` grow for the life of the object, and `session_usage` deep-copies all of them plus re-reads the store on every access.
- `MainAgentSession.restore()` re-attaches to the latest stored turn even if it was `completed`, so `last_usage` after a restore of an idle session is the previous turn's usage. That matches the docs, but `retry()` after restore will add calls to that completed turn.
- `docs/deepagent_overhead_followups.md` is a 218-line planning note; consider whether it belongs in `docs/` or in an issue.

keceli (approved):
Since `442cb51e`: the author's own merge of `main` (`f524353d`) plus three commits (`adee9fa5`, `e2fb09bc`,
`5b81e0a6`). `mergeable: true`; 43 checks green, `test-extras (3.12)` still running at the time of writing.

Verification (v0.7.0-pinned env, absolute `PYTHONPATH`): `test_usage`, `test_cli_usage`, `test_cli_formatting`,
`test_memory`, `test_codex_model`, `test_deep_agent_cli_review`, `test_recursion_limits`, `test_main_agent_cli`
→ **361 passed**, 4 skipped; the single failure is the pty test this sandbox cannot run. `ruff` clean.

## Verdict: Approve

**Rebase on #246 — done.** The author's merge resolves the `cli/formatting.py` conflict the same way I did
(`git diff` of the two merge trees: one blank line). #246's follow-ups (sanitized/bounded previews, tcflush,
recursion-limit precedence) are now in this branch, and its pty/escape tests run here.

**New commits, checked:**

- `adee9fa5` — `on_llm_error` now reads `response_usage(kwargs["response"])` and keeps the counts of a failed
  call as `complete=False`, unless the Codex adapter already supplied usage. LangChain does pass the partial
  generations on stream failures (`on_llm_error(e, response=LLMResult(generations=generations))` in
  `BaseChatModel`), so the branch is reachable; on non-stream failures the response is empty and the record
  stays `failed` with unknown counts, as before. Tests cover the raw-usage retention and the adapter-precedence case.
- `e2fb09bc` — a failed `store.usage_records()` read marks the summary `partial` (and all-unknown when nothing
  else is available) instead of silently presenting in-memory subtotals as complete;
  `test_usage_read_failure_keeps_subtotals_partial_until_recovery` and
  `test_coverage_read_failure_does_not_hide_durable_counts` cover both directions. `format_token_usage` now prints
  "Tokens: unavailable" whenever all three counts are unknown, with reasons only when it has them.
- `5b81e0a6` — `docs/deepagent_overhead_followups.md` removed; no dangling references in `docs/`, `mkdocs.yml`,
  `README.md` or `src/`. The PR body still names that file under "Document the APIs…" — worth a one-line edit
  before squash/merge so the description matches the tree. The deferred items it listed (callback offloading,
  per-update Codex persistence batching, token budgets) now live only in the PR body; if you want them tracked,
  an issue is the place.

All round-1/round-2 items stand as verified in the previous review.

